PR #5, opened 2025-01-28
Loreified Prototypes

troytroy400:
Replaces every mention of the UNMC, Weston-Yamada, SPP, etc with their lore friendly counterparts. 

Replaces all RMC gun names with the lore friendly counterparts.


troytroy400:
Moving these to their own file paths to avoid conflicts.

PR #6, opened 2025-01-28
Title Screen Loreification

troytroy400:
Removes title screens showing other species and adds one from the Aliens TTRPG.


Peptide90:
Changed the license type from NA to custom to satisfy the RGA schema. Applicable licenses can be found here:
https://github.com/space-wizards/RobustToolbox/blob/master/Schemas/rga_validators.py

Outright removing instead of adding them to an ignoredPrototypes.yml file and adding our own in a new /_AU14/ directory might cause merge conflicts in future. Just for reference.

PR #9, opened 2025-02-04
Lethality increase

neki426:
increased lethality on several base USCMC weapons

increased dmg multipliers and decreased scatter on m41a mk1, mk2, m4spr

increased base dmg for 9mm rounds (base 40 -> 44, ap 25 -> 28) slightly decreased scatter when wielded for m4a3 and mod88


troytroy400:
I'm gonna backburner this until upstream has implemented med or someone brings over shitmed and/or a similar system. We need it to be scaled properly for whatever the system is, so it'll just be redundant to push it now.

troytroy400:
Gonna close this out, its no longer needed and the amount of merge conflicts this is generating is insane.

PR #30, opened 2025-02-10
ports AN/PVR-52E1 scope from PVE

neki426:
title
still needs sprite porting but this is the base code
will be changing this more when i actually figure out if the scopes stats are 1 to 1 to the 2x miniscope from main or not

resolves https://github.com/AU-14/AU-14/issues/27


troytroy400:
Passes the lore and gameplay checks. Just needs code review.

neki426:
turned into draft, need to debug and do sprite work

PR #70, opened 2025-04-26
queen and drone sprite ports

neki426:
title
give upstream xeno ai a stylish cm13 resprite
credits in json.meta


troytroy400:
Setting this as do not merge until we've gone over the legal jargin and such.

PR #159, opened 2025-06-05
california guns + usasf ranks

neki426:
makes prototypes for civilian versions of m16, mar-40 and ar-10

adds lower enlisted and commissioned officer ranks for usasf


troytroy400:
Closing this PR as its stable. I'll take it over sooner or later.

PR #171, opened 2025-06-07
ported barricadeblock component from civ14

Evergreendev12:
Ported bullet blocking from Civ14 as a BarricadeBlock component

Example in sandbags.yml


GitHubUser53123:
link your sources please :pray:

troytroy400:
> link your sources please 🙏

Civ 14 is EXTREMLY unorganized with their PRs so finding the original one is a blind turkey shoot. But you can find their repo here, shouldn't be hard to back track it with a good old ctrl + f. https://github.com/Civ13/Civ14

PR #804, opened 2026-04-23
Fixes tunnels being broken

TheHellFireo:
Previous commit had a fix for round start map placed tunnels

That caused a loop to start each time a xeno spawned or evolved, that loop was ment for round start and it ran everytime 

this should fix the tunnels disseapering randomly now.


troytroy400:
Thanks for the PR!

Issue #561, opened 2026-01-29
Crawl State


troytroy400:
If a player is only 50 damage below critical threshold they're given the ability to crawl in a prone state until they are hit by another round or go farther than -50 into the critical threshold.

troytroy400:
Completed by https://github.com/AU-14/ColonialMarinesUniverse/pull/811

PR #811, opened 2026-04-24
Adds the ability to crawl

TheHellFireo:
#561
You will now be able to crawl veryyy slowly if you have less then -30 damage while in crit

Getting brute or burn damage immediately stops you 
and if you get bleed or asphyxiation or like toxin damage below -30 will also stop you.


TheHellFireo:
o yea you can't interact with anything though but can still crawl out of buckled beds and chairs and stuff however given that ur not handcuffed

TheHellFireo:
Should be good now

PR #822, opened 2026-04-24
Vehicle locks can be broken once the vehicle is below 30% integrity

TheHellFireo:
Locks are broken until the vehicle is back up to 90% integrity

The driver and gunner seats of each vehicle given god mode as well

also comment removal


TheHellFireo:
@troytroy400 wasn't merged correctly

PR #967, opened 2026-05-07
Couple build fails fixed

TheHellFireo:


Resources/Prototypes/_AU14/Roles/Departments/au14_colony_department.yml
maertayn, line 1: don't need to do this.
I added .Where(department => department.Primary && !department.EditorHidden) to the JobTest

TheHellFireo: oh got it

PR #969, opened 2026-05-07
HAZOPS Small Update: MOPP Expanded

johnnny34y:
## About the PR
Adds the M17A4 gas mask, matching CBRN armor, and colored CBRN boots for HAZOPS

## Media
image
(no ingame photos because my IDE is bugged, but it's a simple addition so it shouldn't matter)


johnnny34y:
should make HAZOPS a little less "just the HECU" and a little more 80s aliens

PR #1101, opened 2026-05-24
Buffs Dancer and Base Praetorian

Arcticular1:
## About the PR

Dancer:
Base damage increased by 20 
Dodge now lasts 15 seconds and is 10% faster. (this is a movement increase and phase through entities only, not an actual dodge)
Impale now ignores 30 armor and has a 7 second cooldown.
Base Praetorian:
Acid ball do-after no longer cancels when moving.
## Why / Balance

Dancer is extremely underpowered - it doesnt really benefit at all over anything compared to other castes like vanguard and ravager (even being outclassed by vampire) So this is a buff that should help it.

Base Praetorian gets heavily nerfed due to the fact it cant use it's acid ball on the move - forcing it to sit infront of enemies to use the ball (usually limiting its acid stacking by an extreme amount) 
## Requirements

**Changelog**

:cl:
- tweak: Dancer Impale, base damage and dodge has been buffed!
- tweak: Base Praetorian can now use acid ball while moving!


Evergreendev12:
gem alarm

PR #1117, opened 2026-05-26
Flares stomping, job fixes, shuttle refactor

Evergreendev12:
XIDDY XOID 6777777


maertayn:
Extinguish would be nice too for FoF

https://github.com/AU-14/ColonialMarinesUniverse/blob/master/Content.Server/_RMC14/Atmos/RMCFlammableSystem.cs

PR #1189, opened 2026-06-02
USS Bush Weedable + No Tunnels, etc.

400Ferrets:
replaces the EVIL BAD floor tiles with the visually identical GOOD floor tiles

also adds the schtufff to the area to make it not display as open skies (failRP!)

image
image

was going to add more hijack locations but the bush is itty bitty so I didn't

I am aware the bush is currently being remapped'n'shit, and that it will overwrite this, or something. if the new bush has the same issue i'll just fix it again

this also fixes burrowers being able to escape into space, and make tunnels on the bush.

:cl:
- add: the USS Bush is now weed-able by xenos!


troytroy400:
We switched to the MultiZ version of the bush, gotta port this to that sadly.

PR #1223, opened 2026-06-05
Crusher charge bug fixes

Piticzu:
## About the PR

Rewrites the crusher charge ability to use direct physics impulse instead of the engine's ThrownItemSystem. Adds XenoCrusherChargable pass-through tagging to various props so the crusher plows through them during a charge. Doors, shutters, and select vendors now take structural damage on impact.

## Why / Balance

Crusher charge bug fixes:
- **Wall seam phasing**: Charging between two adjacent wall tiles at an angle would let the crusher phase through and fly off the map, killing them instantly.
- **Marines not getting hit**: The ThrownItemSystem only creates its sensor fixture when an entity has exactly 1 fixture. The crusher has 2+ (mob + mob_collision), so the hit detection fixture was never created. Marines beyond close range were never hit or flung.
- **Close-range charge doing nothing**: A previous workaround added a minimum 2-tile range which never fixed the underlying issue and made point-blank charges fail.
- **Doors not taking damage**: Doors were tagged with `instantDestroy` which caused the raycast to skip them, and the StartCollideEvent could miss them due to auto-open timing.

## Technical details

**Charge movement refactor (XenoChargeSystem.cs)**:
- Replaced `TryThrow()` with direct `ApplyLinearImpulse()` + `SetBodyStatus(InAir)`, following the same pattern as the xeno leap system.
- Hit detection now uses `StartCollideEvent` on `XenoChargingComponent` instead of `ThrowDoHitEvent` from ThrownItemSystem. The entity's own hard fixture handles wall collision, which prevents seam phasing because Box2D generates proper separating impulses for hard contacts.
- Charge lifetime managed by `ChargeEndTime` in `Update()` instead of `ThrownItemComponent.LandTime`.
- Added close-range contact check after impulse (handles entities already touching, e.g. adjacent doors).
- Hit logic extracted to `ProcessChargeHit` method, callable from both `StartCollideEvent` and the contact check.
- `XenoChargingComponent` is now networked for client prediction.
- Removed `ThrownItemSystem`, `FixtureSystem`, and `ThrownItemComponent` query dependencies.
- Target fling (marine being thrown after hit) still uses `TryThrow` since the marine entity has a single fixture.

**PreventCollide handler**: Stripped to only pass through entities explicitly tagged with `InstantDestroy + PassOnDestroy`. Everything else collides normally.

**Raycast**: Extended collision mask to include `MidImpassable` so doors are detected in the charge path.

**YAML prototype changes**:
- Doors, shutters, poddoors: Changed from `instantDestroy: true` to `setDamage: { Structural: 750 }` so they take damage on charge impact.
- Select vendors (coffee, snack, cola, soda, cigarettes, nutrimax, seeds, BODA, Koorlander): Added `setDamage: { Structural: 750 }` individually. SecTech, Riot Tech, SEATech, and all loadout vendors are excluded.
- New `instantDestroy + passOnDestroy` tags added to: filing cabinets, floodlights, post lights, disposal units, portable generators, barrels, closets, hospital dividers, platforms.
- Racks and lockers: Added `passOnDestroy: true` to existing `instantDestroy` tags so the crusher passes through instead of stopping.

## Media

N/A, bug fix.

## Requirements

**Changelog**

:cl:
- add: Crusher charge now plays a windup sound during the 0.6s delay before charging.
- fix: Fixed crusher charge phasing through walls when aimed at the seam between two tiles.
- fix: Fixed crusher charge not hitting or flinging marines at range.
- fix: Fixed crusher charge not damaging doors on impact.
- tweak: Crusher charge now plows through props (closets, lockers, racks, barrels, filing cabinets, floodlights, disposal units, generators, platforms, hospital dividers).
- tweak: Doors, shutters, and non-loadout vendors now take structural damage from crusher charge.
- code: Crusher charge rewritten to use direct physics impulse instead of ThrownItemSystem, matching the xeno leap pattern.


Piticzu:
Build has errors that need fixing before I can open it for review again.

Piticzu:
Fixed a YAML issue from the vendor charge damage commit where XenoCrusherChargable was added as a separate components: block on some vendors instead of merging it into the existing one, causing a prototype load crash.

Piticzu:
also added crusher charge-up audio sound

maertayn:
```[DataField] public float? ThrowRange;```
Seems like this is now dead code for prototypes, right?

Piticzu:
You are right, it slipped by me.

Piticzu:
Removed the dead code that remained from the old charge logic

PR #1285, opened 2026-06-10
Revert "Move air deployable sentry from fabricator to ASRS ($600)"

KeiKiyomi:
Reverts AU-14/ColonialMarinesUniverse#1260

Why are we nerfing things that are already unused because of how bad they are. Please show me ONE CLIP of these turreys actually doing something meaingful and not just sitting there rotting because they werent setup. On HvX which is the MAIN PLACE these are used xenos can just, walk around it, or send a DRONE YES A DRONE to destroy it.

Changelog

🆑

tweak: Dropship sentries are NO LONGER in the govfor ASRS for $600 and have been returned to the dropship fabricator.


troytroy400:
Free sentry turrets are not intended.

PR #1305, opened 2026-06-13
HAZOPS Specweapons

johnnny34y:
## About the PR
Adds specweapon kits for HAZOPS. 
Shotgunner kit, which contains the SPAS-12-HAZ and an SOF coif that provides protection
Flamer kit, which contains the M240A2 and the new M17-FLAME which provides fire resistance

**Changelog**
:cl:
- add: Added weapon specialist kits for HAZOPS


Resources/Prototypes/_AU14/Entities/Clothing/HAZOPS/mask.yml
```diff
@@ -43,3 +43,28 @@
     - GasMask
   - type: CMArmor
     bio: 20
+
+  - type: entity
+  parent: [CMBaseMask, CMUBaseGasMaskFilterSlot]
```
maertayn:
```suggestion
  parent: AU14MaskGasHAZOPSM17
```

Resources/Prototypes/_AU14/Entities/Clothing/HAZOPS/mask.yml
```diff
@@ -43,3 +43,28 @@
     - GasMask
   - type: CMArmor
     bio: 20
+
+  - type: entity
```
maertayn:
```suggestion
- type: entity

```

maertayn (changes requested):

Resources/Prototypes/_AU14/Entities/Clothing/HAZOPS/mask.yml
maertayn: Remove the same components, tidy it up a little bit.

HideLayerClothing, Item, Tag & CMArmor doesn't need to be duplicated again - it already has it when doing the above

johnnny34y:
@maertayn changes applied

PR #1343, opened 2026-06-16
Despoiler Embrace QOL + Hypertension Armor mechanic + other fixes

deepdarkdepths:
## About the PR
This PR is a minor rework so this caste can get to a playable state instead of being considered a throw pick.

Despoiler now has armor stacks that goes alongside its hypertension stack mechanic. See Balance section for more details.

Caustic Embrace's leap range has been increased, alongside getting a preview overlay to see where you're actually aiming + where your acid will land. This changes depending on whether it's empowered or not. 

Acid Barrage has a more uniform scatter now, but still has a random chance of where the projectile and its lingering acid will land. The last iteration felt unintuitive to use, but I'll still be looking for feedback on this with the randomized landing.

Oozing Wounds dynamically uses hypertension stacks. The more hurt you are, the less you use, maintaining your armor. If you are full health, all of your stacks are consumed. This better encourages stacking to deal more damage and keeping your survivability with OW, as it was functionally a suicide ability otherwise.

Catalyze now takes 2 hypertension stacks instead of 1, to compensate for previous buffs. Damage bonus per stack reduced, but still higher than initially.

## Why / Balance
Despoiler's buffs weren't enough on its own, so instead of continuing to fiddle with values, I added onto its existing hypertension mechanic to directly help it with surviving engagements. For every 2 stacks it gets from hypertension, its armor increases by 5. At full stacks, it has 40 armor. 
This allows it to better utilize its Oozing Wounds ability, which is based off how low its hp is. This also lets it use the ability without basically needing to kill itself to do so, as Oozing Wounds is quite negligible at high hp. See above for OW changes.

Caustic Embrace's range has been increased from literally only 1 tile to 3 tiles to make it easier and less annoying to use. Empowered embrace retains its 5 tile leap.

## Technical details
New `XenoDespoilerCausticEmbracingComponent` will be used for smoother leaps in the future instead of the current byond port "teleport", currently unused(out of scope for this PR)
`XenoDespoilerCausticEmbraceSystem` uses `direction.Normalized() * NormalRange` landing calculation instead of `SnapDirectionToTile()`. Splash tiles also use `direction` instead of snap tiles. Will likely also be changed to use actual leap logic later

## Media
todo

## Requirements

**Changelog**

:cl:
- add: Despoiler's Hypertension now utilizes armor to be better effective in combat. 
- add: Caustic Embrace now has a preview to see where you're aiming, alongside a range buff.
- tweak: Acid Barrage scatter is more consistent.
- tweak: Catalyze takes 2 stacks to empower, instead of 1. 
- tweak: Oozing Wounds stack use dynamically changes depending on HP; the lower hp you are, the less stacks you consume.


deepdarkdepths:
Check fail unrelated i think

Content.Shared/_RMC14/Xenonids/Despoiler/XenoDespoilerHypertensionSystem.cs
```diff
@@ -19,14 +41,16 @@ public void AddPoints(EntityUid uid, XenoDespoilerHypertensionComponent comp, fl
...
+
         while (comp.Points >= comp.PointsPerStack && comp.Stacks < comp.MaxStacks)
         {
             comp.Points -= comp.PointsPerStack;
             comp.Stacks++;
         }
 
-        if (comp.Stacks >= comp.MaxStacks)
-            comp.Points = 0;
+        if (oldStacks != comp.MaxStacks)
```
TheHellFireo, line 55: This seems to accidentally remove the old max-stack point clamp, despoiler can keep banking Points while already at max stacks, so armor/stacks decay much later than before.

Content.Server/_RMC14/Xenonids/Despoiler/XenoDespoilerOozingWoundsSystem.cs
```diff
@@ -45,9 +47,30 @@ private void OnUse(EntityUid uid, XenoDespoilerComponent comp, XenoDespoilerOozi
...
             return;
 
+        var severity = ComputeSeverity(uid, action);
+
+        if (TryComp<XenoDespoilerHypertensionComponent>(uid, out var hypertension))
+        {
+
+            var stacksToRemove = severity switch
+            {
+                0 => 3,
```
TheHellFireo: you said full-health Oozing Wounds consumes all stacks, but severity 0 removes 3 while Despoiler has 4. Should this remove hypertension instead of hardcoding 3?

Content.Shared/_RMC14/Xenonids/Despoiler/XenoDespoilerCausticEmbracingComponent.cs
```diff
@@ -0,0 +1,24 @@
+﻿using System.Numerics;
+using Robust.Shared.GameStates;
+using Robust.Shared.Map;
+
+namespace Content.Shared._RMC14.Xenonids.Despoiler;
+
+[RegisterComponent]
```
TheHellFireo: This component is currently unused, we can leave this out

Content.Server/_RMC14/Xenonids/Despoiler/XenoDespoilerOozingWoundsSystem.cs
```diff
@@ -45,9 +47,30 @@ private void OnUse(EntityUid uid, XenoDespoilerComponent comp, XenoDespoilerOozi
...
+                1 => 2,
+                2 => 1,
+                _ => 3
+            };
+
+            hypertension.Stacks =
```
TheHellFireo: this is mutating Hypertension stacks directly then refreshing armor manually i'd move that into XenoDespoilerHypertensionSystem, have it handle the stack changes + armor refresh through a helper, keeps everything in one place and stops duplicating the server-side logic.

Content.Client/_RMC14/Xenonids/Targeting/XenoAbilityPreviewOverlay.cs
```diff
@@ -458,6 +476,94 @@ private void DrawDirectionalStomp(
...
     }
 
+    private void DrawCausticEmbrace(
+        in OverlayDrawArgs args,
+        MapCoordinates originMap,
+        MapCoordinates mousePos,
+        XenoDespoilerCausticEmbraceActionComponent embrace,
+        EntityUid player)
+    {
+        {
```
TheHellFireo: Remove the extra braces to keep it consistent with the rest of this overlay

Content.Server/_RMC14/Xenonids/Despoiler/XenoDespoilerCausticEmbraceSystem.cs
TheHellFireo: forward is now a normalized direction so we can change this to

private static Vector2i SnapDirectionToTile(Vector2 dir)
{
    return new Vector2i(Math.Sign(MathF.Round(dir.X)), Math.Sign(MathF.Round(dir.Y)));
}

TheHellFireo: we can then remove these to be just
var back = SnapDirectionToTile(-forward);

TheHellFireo: then these can become back.X and back.Y

Content.Shared/_RMC14/Xenonids/Despoiler/XenoDespoilerCausticEmbracingComponent.cs
deepdarkdepths: I'll just delete it since it's just as easy to add it back

PR #1347, opened 2026-06-17
Yautja flame resist (35%)

Arcticular1:
## About the PR

Yautja are now 35% flame resistant 
## Why / Balance

Same as CM (preds right now get 1 tap crit by x napalm, so this will allow them to roll and MAYBE not die)
## Media

None
## Requirements

**Changelog**

:cl:
- tweak: Predators now have a 35% flame resist.


maertayn:
It's not used by anything

Arcticular1:
> It's not used by anything

Idk this means

maertayn:
You've attached a value, but you're not using the float anywhere. So it's just dead code / unused data

You'll need to find where fire damage is applied, and then if it can find the component (null check), multiply that value with the fireDamageMultiplier, before returning the final damage

Arcticular1:
> You've attached a value, but you're not using the float anywhere. So it's just dead code / unused data
> 
> You'll need to find where fire damage is applied, and then if it can find the component (null check), multiply that value with the fireDamageMultiplier, before returning the final damage

all done

maertayn:
Would've preferred seeing the damageCoefficient instead though.

PR #1355, opened 2026-06-17
Adds ChemicalIrritant system for human gas chemicals

HotdogMan52:
## About the PR
Allo, i've always wanted to use tear gas without being affected while wearing gas masks (i think it'd be so cool to throw gas at rioters and then emerge past it without care). With that in mind, I spent like a week or so studying and trying to create a system for it (also lowkey praying it actually works outside of a localhost dev env...)

## Why / Balance
Right now, tear gas and CN20 can be used to force crowds to disperse but is a double-edged blade because gas masks don't do anything. With this, it'll allow individuals with gas masks AND filters to just walk through the gas without being debuffed (albeit filter takes damage) so GOVFOR can march past tear gas to arrest anyone OR CLF can toss CN20 and use it as hit-and-run cover while wearing gas masks against unprepared GOVFOR

## Technical details
added 4 files to Content.Shared\_CMU14\Weapons\ChemicalIrritants\

-ChemicalIrritantProfile.cs which is a list of settings that describe how an gas behaves

-ChemicalIrritantInjectorComponent.cs which is attached to the gas cloud, tracks who it's touching and how much gas it has left (planning to use this for pepper spray or smthn)

-ChemicalIrritantComponent.cs attached to the victim, tracks how much gas they've absorbed.

-SharedChemicalIrritantSystem.cs which is a fork(?) of the neurotoxin system used by boiler's neuro gas but with different effects and modularity

Added 1 file to Content.Server\_CMU14\Weapons

-ChemicalIrritantSystem.cs needed to actually run the chemicalirritant system

Tweaked 1 file
teargassmoke.yml to work with new system instead of the xeno neurotoxin

Additionally, the system is modular so you can now make new gases with varying intensity of effects through YAML only 

## Media
https://youtu.be/m2NbZMKTWM0

## Requirements

**Changelog**
:cl:
- add: Tear gas now respects gas masks with filters. Wearing a gas mask with a filter lets you walk through chemical gas unaffected, though the filter takes damage.
- tweak: Tear gas smoke now uses the new ChemicalIrritant system instead of the xeno neurotoxin system.

### DIFF
--- a/Content.Server/_CMU14/Weapons/ChemicalIrritantSystem.cs
+++ b/Content.Server/_CMU14/Weapons/ChemicalIrritantSystem.cs
@@ -1,0 +1,16 @@
+using Content.Shared._CMU14.ChemicalIrritants;
+
+namespace Content.Server._CMU14.Weapons;
+
+/// <summary>
+/// Server-side system for chemical irritants (tear gas, pepper spray, etc.)
+/// This registers and runs the shared system on the server.
+/// </summary>
+public sealed class ChemicalIrritantSystem : SharedChemicalIrritantSystem
+{
+    public override void Initialize()
+    {
+        base.Initialize();
+        // Any server-specific logic can go here
+    }
+}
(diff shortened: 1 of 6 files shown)


HotdogMan52:
holy crap checks didn't fail horribly

image

PR #1360, opened 2026-06-18
M-19 "Treadnailer" assault boots, UZI Tweak

wray-git:
## About the PR
# Revert PR if there is a new unexplainable lag. Uses a tick system.
This adds the new M-19 "Treadnailer" boots. It hurts anything you walk above
- Humans
- Xenos
- Resin

The Treadnailer is only available in the **Radio Operator equipment rack** since its the closest thing to a Squad Leader vendor. It originally is meant to only be available to _**Squad leaders**_(where theres only 3 in the round.) HAZOPs get an additional 3 Treadnailers in their requisition vendor.

Also, Uzi now has a higher RoF, more accurate, can be stored in pockets, belt, inside backpack, and as backpack slot.
Bug: Just needs to be sprited for those slots.
Potential Change: Make Uzi not storeable in pockets.

Possible Bug: Range of the Treadnailer (possibly hitting targets slightly outside of the tile they occupy)

Possible future change: _Increase damage against humans if SLs keep successfully avoiding hitting their teammates_

## Why / Balance
**It has more cons than pros**(as I intended with all my new upcoming features):
🔸 Has an intense 20% slowdown(may lower in the future, may have to make it toggleable with long cooldown in the future due to its punishing slowdown)
🔸 Make a loud, unique audible noise (open for anyone to change it). This is to *alert any hostiles*.
🔸 Damages both marines and xenos
🔸 ~~Only 3 in the round, given to SL vendors~~. Added to Radio Telephone Operator Vendor(Restricted to Lead roles). HAZOPs get 3 extra in Requesitions

🔹 Deals 10 pierce to humans(2.0s cooldown)
🔹 Deals 60 pierce to xenos (0.75s cooldown)
       ⚠️ 🔸 **Intense screenshake when running over a player(human or xeno) to make it difficult to aim**
🔹Breaks resin

- Combatting queen's eye kind of broken resin building(due to plasma tree spamming, queens have infinite plasma)
- Gives a better chance for marines to push due to:
- - Variety of maps are harder to balance, thus resulting in chokeholds(most rounds end with marines losing in the same chokeholds)
- - Variety of maps are harder to balance, thus resulting in larger than usual maps in which xenos are able to easily flank due to certain maps allowing it.

For UZI:
- da UZI was not UZIing
- Its RoF was increased to `7.5`
- Its accuracy is slightly improved, but it still is widely inaccurate. I'd say it shouldn't get more buffs because it's practically the better alternative for the 1911 because it allows for those quick kills, with a larger magazine(1911's mag barely lasts 1 kill), **but** (still) extremely inaccurate.

## Technical details

The core of this is a new server-side EntitySystem (`SpikeBootsSystem`) that runs a tile-change check every update tick. Rather than doing expensive work every frame, it compares the wearer's current tile position (`Vector2i`) against the last recorded tile and only fires logic on actual movement to a new tile. On tile entry it queries all entities on that tile via `EntityLookupSystem.GetLocalEntitiesIntersecting`, separates resin entities from mobs, and handles each category independently.

Resin destruction just queues a delete. Mob damage goes through a passability check that covers three separate cases: `StandingStateSystem.IsDown` for prone/crit/unconscious mobs, `XenoRestingComponent` for resting xenos which cancel their own collision events independently of standing state, and a dedicated `XenoParasiteComponent` check for parasites which use `SmallMobLayer`/`SmallMobMask` physics layers that don't interact with bipedal mob layers at all, making them physically passable without being "down" by any other measure.

Damage is split by xeno vs non-xeno with separate cooldowns (0.75s xenos, 2s marines) tracked in a per-wearer dictionary on `SpikeBootsWearerComponent`. Cooldown cleanup runs on tile change only, using a system-level `List<EntityUid>` to avoid heap allocation.

Speed slowdown is handled entirely by `ClothingSpeedModifierComponent` in YAML (the standard armor pattern), with explicit `RefreshMovementSpeedModifiers` calls on equip and unequip to trigger the inventory relay recalculation, which does not happen automatically.

### M-19 "Treadnailer" in action
Content Client_Tjt09Rmek7

https://github.com/user-attachments/assets/231c41fa-638a-46a7-b159-196b187dcc5e

Uzi before:
https://github.com/user-attachments/assets/359f61e3-3a33-4766-9443-a0bc50e2f4f0
Uzi after:
https://github.com/user-attachments/assets/51e48dac-6971-4635-8f1b-16091ea4c63b

## Requirements

**Changelog**

:cl:
- add: Added the M-19 "Treadnailer" assault boots, a special issue piece of kit fitted with tungsten-carbide spike plates in the sole. Destroys xeno resin on contact, and if you happen to walk over anyone who's in no shape to get out of the way, they're gonna have a bad time. Slowdown 20%. Available from RTO(lead) lockers, HAZOPs get 3 from their cargo vendor under "Special Issue Equipment"
- add: UZI can now be stored in bags, belts, pockets, and backpack slot
- tweak: Changed HAZOPs' cargo vendor section from "Special Issue Firearms" to "Special Issue Equipment"
- tweak: Increased UZI RoF to 7.5
- tweak: Improved UZI accuracy slightly 🤏 
- code: Added SpikeBootsComponent, a marker component that sits on the boots item and does nothing exciting on its own
- code: Added SpikeBootsWearerComponent, slapped onto whoever equips the boots, keeps track of what tile they last stepped on and who they recently stomped so we don't spam damage every tick
- code: Added SpikeBootsSystem, the thing that actually does stuff, runs a tile-change check each update and handles resin destruction, passable mob damage, cooldowns, sounds and screenshake
- code: Added spikedboots.rsi to _AU14/Clothing/Shoes with the full icon/equipped/filled state set


wray-git:
@maertayn 
any issues? I don't know RMC's tests 🫠

KeiKiyomi:
Test fails are fake, it's yapping about humvee and vehicle load outs

PR #1361, opened 2026-06-18
Replaces M240-T with M240 + purchasable napalm

Arcticular1:
## About the PR

Replaces the M240-T in ASRS with two M240 flamers, as well as adds purchasable napalm.
2 M240 - 800
4 UT-NAPALM: 200
2/2 UT/BGEL: 350
4 NAPALM: 500

## Why / Balance

The T variant should stay a specialist weapon with its added armor, as well as having more flamers on the field is always good, rather than having to rely on a singular person (two for the price of 1).
You can still get OT chemicals for that extra kick.
Being able to refuel on napalm from ASRS rather than having the limited standard supply.
## Technical details

YAML
## Requirements

**Changelog**

:cl:
- add: Added 2x M240 Flamer (800$) to ASRS
- add: Added 4x UT Fuel (200$) to ASRS
- add: Added 2x UT Fuel 2x BGEL Fuel mixed crate (350$) to ASRS
- add: Added 4x BGEL Fuel (500$) to ASRS
- remove: Removed M240-T from asrs.


Arcticular1:
I’ll fix this later

mikusssssss:
Killing my baby, you monster.

OmerF-afk:
Instead of removing M240-T entirely, why not keep it as an expensive purchase instead? Trying to make every weapon become part of the WS system is not only boring, but it also takes away from the GOVFOR's ability to buy whatever is needed for the mission, since only 4 people can use it

I feel like this will just make people go for SMAW instead, and a week later, people will make a PR about making SMAW spec-only as well. Which is boring again.

Arcticular1:
> Instead of removing M240-T entirely, why not keep it as an expensive purchase instead? Trying to make every weapon become part of the WS system is not only boring, but it also takes away from the GOVFOR's ability to buy whatever is needed for the mission, since only 4 people can use it
> 
> I feel like this will just make people go for SMAW instead, and a week later, people will make a PR about making SMAW spec-only as well. Which is boring again.

Because the m240-t would be better with a dedicated specialist for it - it has the fireproof armor, it could have more options and etc rather than being a standard flamer. If we gave this the ability to have x napalm what would the difference be? Its really only the fuel capacity from what i saw.

PR #1399, opened 2026-06-20
CO's Silence Order

wray-git:
## About the PR

Commanding Officers are now able to issue a Silence order, an order that would force marines around them to only whisper for 10 seconds, with a 10 second cooldown.

**⚠️ The action icon needs a sprite! A xeno sprite has been used as a placeholder!**
## Why / Balance

For hopefully, more organized briefings
(this will definitely not backlash with spam-pointing)
## Technical details

New `Silence Order` ability for the GOVFOR Platoon Commander. It runs completely separate from the existing Move/Focus/Hold order system so it has its own cooldown and doesn't interfere with those at all.

**What's new:**
- `AU14SilenceOrderAbilityComponent` sits on the CO (added via job `AddComponentSpecial`). It stores the action ref, a 10s cooldown, 10s effect duration and a 7 tile range.
- `AU14SilenceOrderComponent` is a networked component that gets slapped onto affected marines for the duration. It just tracks an `ExpiresAt` time and gets cleaned up by the system's `Update` loop when it runs out.

**Server system (`AU14SilenceOrderSystem`):**

Registers the action on startup, removes it on shutdown. When the action fires it sets its own use delay (does NOT touch Move/Focus/Hold cooldowns), grabs all `MarineComponent` entities in range, applies `AU14SilenceOrderComponent` to each one (reuse refreshes the timer), sends each target a private chat notice and popup, then sends the CO's throat-clear emote. The CO is skipped from the silence effect. Pronoun in the emote (his/her/their) is pulled from `HumanoidAppearanceComponent.Sex`. The emote bypasses the normal `SendEntityEmote` path so we can control font size and color directly via `SharedCMChatSystem.ChatMessageToMany` with a hand-built `wrappedMessage`.

**Chat hook (`ChatSystem.TrySendInGameICMessage`):**

Two lines added just before the `InGameICChatType` switch. If the message type is `Speak` and the source has `AU14SilenceOrderComponent`, it gets quietly redirected to `Whisper`. Simple and clean, works across all normal speech paths.

## Media

https://github.com/user-attachments/assets/b1b7f3d4-f69e-4d5c-8f65-6e44d2a36bbc

Discord_88mjLUneY8

## Requirements

**Changelog**

:cl:
- add: GOVFOR Platoon Commanders have a new Silence Order ability. Activating it forces nearby marines to whisper only for 10 seconds and sends them a subtle notification. The CO emotes a loud throat clear instead of shouting a callout. Has its own 10 second cooldown separate from the other orders.

### DIFF
--- a/Content.Server/Chat/Systems/ChatSystem.cs
+++ b/Content.Server/Chat/Systems/ChatSystem.cs
@@ -18,6 +18,7 @@ using Content.Shared._CMU14.Yautja;
 using Content.Shared._RMC14.CCVar;
 using Content.Shared._RMC14.Chat;
 using Content.Shared._RMC14.Mentor.ImaginaryFriend;
+using Content.Shared._AU14.Marines.Orders;
 using Content.Shared._RMC14.Stun;
 using Content.Shared._RMC14.Xenonids;
 using Content.Shared.ActionBlocker;
@@ -292,6 +293,10 @@ public sealed partial class ChatSystem : SharedChatSystem
                 return;
             }
         }
+
+        // AU14: Silence Order forces speak → whisper for the duration.
+        if (desiredType == InGameICChatType.Speak && HasComp<AU14SilenceOrderComponent>(source))
+            desiredType = InGameICChatType.Whisper;
 
         // Otherwise, send whatever type.
         switch (desiredType)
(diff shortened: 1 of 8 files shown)


maertayn:
Duration should probably be longer than 10 seconds to be effective for the use-case, if it's meant to be used during briefings (~half of the duration is used up on typing). Cooldown is fine, as long as it doesn't get abused. In current implementation it's more a reminder, not enforcement, to quiet down which is more considerate (just hope people don't switch to looc rightaway)

Pretty cool.

wray-git:
@maertayn I actually wanted to disable LOOC, but chose not to.
Currently aye, it's meant to be as a reminder to be quiet. Most of the PRs you'll see coming from me often "hit weak" first with variables that can easily edit their functionality with simple value changes.

I invite you to edit it if you'd like, and especially change the sprite of the action icon.

PR #1414, opened 2026-06-22
Create weapon sling system

HotdogMan52:
## About the PR
based off magnetic harness but without the magnets part and reliance on armor
makes sling act like a sling

## Why / Balance
just found it funny how a sling which wraps around your body doesn't work just because you don't have armor or your armor isn't magnetic

now, it's actually going to make sure your item is slung (on your back or on your armor/outer clothing)

idk how this'll affect balance lowkey so probably gonna need it to go around a few rounds to see if it actually changes balance

## Technical details
c sharp files which made me pull on my hair because i couldn't fucking understand the magharness system
i just lowkey copied magharneess but compressed as much as i can 

yml to make sling use new stuff and make new sling standard

ftl for popup messages

## Media
https://youtu.be/Qq1XkP9ZMNU

## Requirements

**Changelog**
:cl:
- tweak: Changed two-point slings to not rely on armor like magnetic harnesses
- tweak: Replaced all RMCAttachmentTwoPointSling with the AU14AttachmentTwoPointSling

### DIFF
--- a/Content.Shared/_CMU14/Attachable/Components/AttachableSlingComponent.cs
+++ b/Content.Shared/_CMU14/Attachable/Components/AttachableSlingComponent.cs
@@ -1,0 +1,5 @@
+namespace Content.Shared._CMU14.Attachable;
+
+[RegisterComponent]
+[Access(typeof(CMUSlingSystem))]
+public sealed partial class AttachableSlingComponent : Component;
--- a/Content.Shared/_CMU14/Attachable/Components/CMUSlingItemComponent.cs
+++ b/Content.Shared/_CMU14/Attachable/Components/CMUSlingItemComponent.cs
@@ -1,0 +1,7 @@
+using Robust.Shared.GameStates;
+
+namespace Content.Shared._CMU14.Attachable;
+
+[RegisterComponent, NetworkedComponent]
+[Access(typeof(CMUSlingSystem))]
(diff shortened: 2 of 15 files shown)


HotdogMan52:
Failed PlanetLoadsWithoutErrorsTest [2 m 28 s]
  Error Message:
     Failed to load planet JE-1758 "Ascanius"!
     
dunno if i messed up anything else

maertayn:
Why not edit the current magnetic harness system instead? Making a whole new system to make one or the other redundant isn't the play here

HotdogMan52:
> Why not edit the current magnetic harness system instead? Making a whole new system to make one or the other redundant isn't the play here

Uh good point actually yea

I'll do that
Are we planning on keeping/using the mag harnesses(?)

maertayn:
Okay, so I looked at it. It specifically requires `RMCMagneticArmor` component which light armor lacks.

If you can add a boolean in the `AttachableMagneticComponent` comp that is something like `NeedsMagneticField`, then look at `RMCMagneticSystem.CanReturn`, add the check in there, something like this

if (TryComp<AttachableMagneticComponent>(ent, out var mag) && !mag.NeedsMagneticField) magnetizer = user; return true;

That way you don't overengineer stuff and it's just an easy fix without having to add RMCMagneticArmor to every prototype, note this will allow the sling to be used if you only have a suitstorage slot - just like you intended

PR #1430, opened 2026-06-24
Add's Plant Bags Into the Hydroponics Crate

OverusedMeme:
## About the PR
Added Two Plant Bags to the hydroponics crate

## Why / Balance
I saw a feedback post on it, found it odd that the crate that has  hydro gear except plant bags, which make botany less irksome to do. Made it a total of two since one bag fills rather quickly and two makes trips more worth while. 

**Changelog**
:cl:
- add: Added two plant bags to the hydroponics crate


deepdarkdepths:
`AU14CrateHydroponics storage fill is too large.
Assert.That(size, Is.LessThanOrEqualTo(entStorage.Capacity))
  Expected: less than or equal to 30
  But was:  32`

Either remove something from the crate or make it its own crate. It might just spawn on top of the crate in-game so idk how much you have to worry about this. Test it

maertayn:
Holy shit i'm an idiot, it was 2 fuxking lines

PR #1459, opened 2026-06-26
15$ Per CLF Point

Arcticular1:
## About the PR

Halves the CLF point cost
## Why / Balance

Economy has been crippled in a sense compared to when this was in effect, including some price increases in vendors and etc. This addition would be good, too.

Currently the gameplay loop makes CLF makes CLF calling third parties and etc unreasonable. This would give them an edge without improving their equipment by giving them more options to choose from in their current selection.
## Requirements

**Changelog**

:cl:
- tweak: Clf now get 1 point per 15$

### DIFF
--- a/Content.Server/AU14/AnalyzerSystem.cs
+++ b/Content.Server/AU14/AnalyzerSystem.cs
@@ -21,7 +21,7 @@ public sealed partial class AnalyzerSystem : EntitySystem
     [Dependency] private PopupSystem _popupSystem = default!;
 
     private const string ClfFaction = "clf";
-    private const int CashPerPoint = 30;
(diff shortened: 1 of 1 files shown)


maertayn:
I don't play CLF, so I cannot judge how bad it is, but ideally we fix the economy issues (which you mentioned as being cripled) and not hacky fix it like this

Arcticular1:
> I don't play CLF, so I cannot judge how bad it is, but ideally we fix the economy issues (which you mentioned as being cripled) and not hacky fix it like this

Fair enough. This was just a solution to that because I don’t know what’s intended for the colony specifically.

maertayn:
real, alright we can always change the hacky fix back later.

PR #1461, opened 2026-06-26
Pyro specialist buff.

Arcticular1:
## About the PR

Gave the armor light speed and pounce protection.
## Why / Balance

Reduces the chance to get captured while using the armor in an intended way, as it will cause people to skip out on using the armor.
## Requirements

**Changelog**

:cl:
- tweak: Pyro armor is now immune to pounces, and is now light armor!

### DIFF
--- a/Resources/Prototypes/_RMC14/Entities/Clothing/OuterClothing/marine_armor.yml
+++ b/Resources/Prototypes/_RMC14/Entities/Clothing/OuterClothing/marine_armor.yml
@@ -797,6 +797,11 @@
     isCorrodible: false
   - type: CMArmor
     bio: 25
+  - type: RMCArmorSpeedTier
+    speedTier: light
+  - type: ClothingSpeedModifier
+    walkModifier: 0.725
+    sprintModifier: 0.725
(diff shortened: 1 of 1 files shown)


maertayn:
Why did the M3-G4 grenadier armor get forward-facing `RMCGrantLeapProtection`?

I know you didn't do that, but I'm curious what the idea behind it was- and by extension why the pyro should be getting the same protection. I didn't even know this was a thing tbh

Arcticular1:
> Why did the M3-G4 grenadier armor get forward-facing `RMCGrantLeapProtection`?
> 
> I know you didn't do that, but I'm curious what the idea behind it was- and by extension why the pyro should be getting the same protection. I didn't even know this was a thing tbh

So the idea is that the pyrospecialist is usually on the front when using their abilities -> standing in fire and etc. I believe this would help stop xenos from grabbing them when do their firewalks (armour wasn’t ever taken due to xenos being able to grab them while they were standing in fire as well as the caveat of it being slow).

maertayn:
solid.

PR #1468, opened 2026-06-27
Warlock and Reaper buffs

deepdarkdepths:
## About the PR
Warlock and Reaper have had their ability values buffed up from what they were previously.
Both T3 castes were severely lacking in performance, and were considered noob traps/throw picks. 

Reaper:
Rapture gives 60 flesh resin, up from 30
Flesh Bloom's cost is 100, down from 120
Bloom heals for 240, up from 65.
Bloom's damage is 210, up from 120
Red Gas's range is 8, up from 7
Gas does 180 toxin, up from 120.
Gas costs 20 resin, up from 10.
Carrion Mantle's cost is 120, down from 150
Mantle's shield decay is 80, down from 100

Warlock:
Psychic Crush damage is 65, up from 30
Crush dmg per pulse is 35, up from 10
Psychic Blast's damage is 55, up from 35
Psychic Shield max projectile block is 10, up from 8

## Why / Balance
Reaper's flesh resin concept made it awkward to work with despite no longer passively decaying, and its gas and bloom did not have a meaningful impact. For being a support T3, it could not do that very well if at all. The only thing that was remotely good was the shield and building, but that by itself wasn't enough of a reason to pick Reaper. 
Warlock's abilities simply did not do enough damage, especially Psychic Crush. The long windup for maximum damage is now rewarded.

As a note, Reaper's damage values here are finicky as both the gas and bloom are inconsistent in where they apply the most damage, so this is only listing the maximum damage you can take while holding still.

## Technical details
Reaper has had its component modified as a .yml file doesn't exist for it for some reason.
Warlock's system file has also been directly modified as it lacks both a proper component, with only one .yml file that allows changing of psychic blast's damage, with the rest being in system.

## Requirements

**Changelog**
:cl:
- tweak: Reaper's values for its Rapture ability, Bloom ability, Gas ability, and Mantle have been increased to better its viability.
- tweak: Warlock's damage has been increased for all of its abilities, with a slight buff to its shield bullet cap.

### DIFF
--- a/Content.Shared/_CMU14/Threats/Mobs/Xeno/Caste/Warlock/CMUXenoWarlockSystem.cs
+++ b/Content.Shared/_CMU14/Threats/Mobs/Xeno/Caste/Warlock/CMUXenoWarlockSystem.cs
@@ -452,7 +452,7 @@ public sealed partial class CMUXenoWarlockSystem : EntitySystem
     public const float PsychicShieldHalfThickness = 0.5f;
     public const float PsychicShieldHalfWidth = 1.5f;
     public const float PsychicShieldProjectileStopOffset = 0.1f;
-    public const float PsychicShieldReflectionSpreadDegrees = 50f;
+    public const float PsychicShieldReflectionSpreadDegrees = 80f;
 
     private const float WarlockDirectedParticleVelocity = 7f;
     private readonly HashSet<EntityUid> _affected = new();
--- a/Content.Shared/_CMU14/Threats/Mobs/Xeno/Caste/Warlock/CMUXenoWarlockSystem.cs
+++ b/Content.Shared/_CMU14/Threats/Mobs/Xeno/Caste/Warlock/CMUXenoWarlockSystem.cs
@@ -439,20 +439,20 @@ public sealed partial class CMUXenoWarlockSystem : EntitySystem
     private static readonly FixedPoint2 PsychicCrushMechDamageMultiplier = FixedPoint2.New(2.3f);
     public const string PsychicBlastFireSoundPath = "/Audio/_CMU14/Xeno/Warlock/volkite_4.ogg";
     public const string PsychicBlastImpactSoundPath = "/Audio/_CMU14/Xeno/Warlock/EMPulse.ogg";
-    public const int PsychicCrushBaseDamage = 30;
-    public const int PsychicCrushDamagePerPulse = 10;
+    public const int PsychicCrushBaseDamage = 65;
+    public const int PsychicCrushDamagePerPulse = 35;
     public const int PsychicCrushMaxAreaRadius = 2;
     public const int PsychicCrushMaxPulses = 5;
     public const int PsychicCrushPlasmaPerPulse = 40;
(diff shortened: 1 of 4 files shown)


deepdarkdepths:
@maertayn Do i gotta worry about these test fails

maertayn:
Haven't gone through all, but unlikely, it's just checks to make sure things are as they should be.

```
Assert.That(CMUXenoWarlockSystem.GetPsychicCrushDamage(-1), Is.EqualTo(30))
  Expected: 30
  But was:  65
```

You probably buffed damage, so then it's like- that's not good (when it is)

deepdarkdepths:
> Haven't gone through all, but unlikely, it's just checks to make sure things are as they should be.
> 
> ```
> Assert.That(CMUXenoWarlockSystem.GetPsychicCrushDamage(-1), Is.EqualTo(30))
>   Expected: 30
>   But was:  65
> ```
> 
> You probably buffed damage, so then it's like- that's not good (when it is)

Should i modify the tests so the fails don't happen anymore? It's just screaming about the value not being correct so idk if that would be a massive problem. Or can it just be skipped

maertayn:
Yeah, give it a shot. It can't be skipped, it would affect all future PRs with the same issues until fixed/removed.

I'm all for making its name match its worth, but I don't know how practical the reaper/warlock is in a live match. As such, I'm only comparing current to new changes & see a x4 increase for healing and new insta-kills (100+ damage based on the test assertion) with a stack of 1? on psychic damage - which can grow to 5 stacks..

It's better if Chip reviewed this, they PRd the Warlock and know what specs would be warranted to buff it

deepdarkdepths:
> Yeah, give it a shot. It can't be skipped, it would affect all future PRs with the same issues until fixed/removed.
> 
> I'm all for making its name match its worth, but I don't know how practical the reaper is in a live match. As such, I'm only comparing current to new changes & see a x4 increase for healing and new insta-kills (100+ damage based on the test assertion) with a stack of 1? on psychic damage - which can grow to 5 stacks..
> 
> It's better if Chip reviewed this, they PRd the Warlock and know what specs would be warranted to buff it

Alright, I'll do that later today.
As for the balance concerns, I've already tested changes for both thoroughly. Combined with both being predictable and damage output not being consistent, I buffed the values for bloom and gas since it didn't really seem to be doing anything to marines that had a normal human reaction time. The gas itself takes some time to get to where it needs to be, and bloom has a decent doafter, with the bloom itself being visible where it's going to hit. The healing was pretty bad- around 65? So I made it actually viable. For reference, a drone has around 400 hp. Though I'll likely increase the resin cost for it now
Warlock's Psychic Crush also takes about.. 4-5 seconds or so to get to its maximum damage state, during which it's very visible and you can only be hit if you are unable to move, or slowed. When the crush is let go early, it does negligible damage. It's a charge-up thing, and is underpowered on live.
Psychic Blast was upped to just be more impactful, as I didn't really see much being done with it. 

I'm fine with Chip looking at this

maertayn:
We'll rock with it and tweak it.

PR #1509, opened 2026-07-01
Doors & Walls changes

HolySSSS:
## Technical details
YMLMaxxing at its best.
**IMPORTANT!: Adds wallTag as a half-measure to Doors. This is NOT a good idea to merge without double checking.**
The wallTag was added so that players who want to weld the door won't have to fix it first.

RMCRepairableComponent is programmed to check only walls before making an alternative verb,
check (CTRL+F) "private void OnRepairableAlternativeVerb" in RMCRepairableComponent.
If someone fixes the wallTag check in the code that will be appreciated.

Otherwise it seems to work as intended, though I did not bug test this meticulously.
(building, using, deconstructing, welding, breaking and repairing doors seems to work fine)

**Changelog**
:cl:
- tweak: Doors are repairable with a welder now.
- fix: Doors and walls cannot be broken with fists and small blunt items now.


HolySSSS:
I messed up the YML somewhere parents, will fix.

HolySSSS:
Gonna atomise this PR into two with fixing doors and the resistance to blunt damage.

PR #1306, opened 2026-06-13
feat(radio): add AN/PRC-117G comms system

rokudara-sen:
## About the PR

This PR adds the AN/PRC-117G manpack radio for Radio Telephone Operators, a tunable direct-frequency layer for all headsets, tiered radio jamming, and an electronic-warfare loop (direction finding, interception, COMSEC changeovers) on top of it.

The goal is to make ground comms more interesting than a simple "tower works / tower does not work" check. RTOs now have a real field role: keeping squads connected, managing secure nets, handling intercepts, watching their emission signature, and reacting to jamming.

### The radio

The AN/PRC-117G is worn on the back and can only be operated by trained roles — for GOVFOR the RTO, Commander, Officer Adjutant, and Advisor; for CLF the Cell Leader. This is deliberate redundancy: if the RTO drops, another trained role can pick up the already-keyed pack and keep the element on the net. It supports up to four programmable preset slots with user-defined labels, assigned either from the faction net list or entered manually as raw frequencies. The active AN/PRC net is transmitted on with `:r`, prefixed by an operator-set callsign.

While worn and powered, the radio acts as a relay anchor. Friendly headset traffic on nets loaded into the AN/PRC works without a communications tower within 30 tiles at full signal, and out to 45 tiles with degraded signal (stock figures — see power and antennas below).

The radio has four operating modes, each with a distinct up and down so none of them is the meta:

* **FH — Frequency Hopping**: jamming against you is reduced by one tier and you can transmit from inside jam fields; costs 1.5× battery and even secured FH traffic carries a small DF signature. Interceptors without matching fill get light hop interference on in-the-clear FH traffic.
* **SC — Single Channel**: the efficient workhorse. 0.75× battery and zero DF signature while secured; no jamming resistance.
* **CT — Cipher Text**: invisible to interceptors (no red text, no log entry) and no DF signature, but only fill-loaded AN/PRCs can decode it — your own riflemen hear an encrypted burst, so it is useless for squad broadcast. Transmit is locked until fill is loaded. No jamming resistance.
* **PT — Plain Text**: the emergency mode. 0.5× battery, but readable by everyone, jammed one tier harder, and the biggest DF signature.

It also has monitor (receive-only) mode, scan mode with automatic active-slot switching, a `0–4` squelch noise gate, and a 50-entry server-side net log that survives closing the panel and powering off. The log can be filtered client-side by net (dropdown) and by free-text search over sender and message.

[...]

PR #1568, opened 2026-07-10
feat(radio): AN/PRC-117G comms system

rokudara-sen:
## About the PR

This PR contains the full AN/PRC-117G communications system in a single change set. It was previously opened as #1306 and was briefly split into a 12-PR stack, following my usual workflow. Based on review feedback, I have recombined it here and closed #1569.

The PR adds the AN/PRC-117G manpack radio for Radio Telephone Operators, direct-frequency tuning for all headsets, tiered radio jamming, and an electronic-warfare loop covering direction finding, interception, and COMSEC changeovers. It also adds a corded handset for command staff, mappable comms arrays and masts, and automatic faction callsigns with a directory console.

The main goal is to move ground communications beyond a binary "tower works / tower does not work" check. RTOs now have an active field role: keeping squads connected, managing secure nets, handling intercepts, controlling their emissions, and responding to jamming.

> [!NOTE]
> **The entire feature is gated behind a new `au14.new_comms_system` CVar, default off.** With the CVar disabled, radio behavior is unchanged. A test merge will not affect the server until the option is enabled. See the merge warning below for the mapping requirements that must be in place first.

### The radio

The AN/PRC-117G is worn in the back slot and can only be operated by trained roles. For GOVFOR, those roles are the RTO, Commander, Officer Adjutant, and Advisor; for CLF, the Cell Leader can operate it. This redundancy is intentional: if the RTO goes down, another trained role can pick up the keyed radio and keep the element connected.

The radio supports up to four programmable presets with user-defined labels. Presets can be selected from the faction net list or entered as raw frequencies. The active AN/PRC net uses `:r`. Transmissions are prefixed with the operator's callsign, which replaces the speaker's name on the radio message. Operators without a callsign appear as "UNKNOWN STATION."

While worn and powered, the radio acts as a relay anchor. Friendly headset traffic on nets loaded into the AN/PRC works without a communications tower within 30 tiles at full signal and up to 45 tiles with degraded signal. These are the stock ranges before TX power and antenna modifiers.

The radio has four operating modes, each with a specific use case and a real downside:

* **FH — Frequency Hopping**: reduces incoming jamming by one tier and allows transmission from inside a jam field. It uses 1.5× battery, and secured FH traffic still carries a small DF signature. Interceptors without the matching fill hear light hop interference when the FH traffic is otherwise in the clear.
* **SC — Single Channel**: the efficient default. It uses 0.75× battery and has no DF signature while secured, but offers no jamming resistance.
* **CT — Cipher Text**: produces no intercept text, no log entry, and no DF signature. Only AN/PRCs with matching fill can decode it; normal squad headsets receive an encrypted burst, so CT is not suitable for general squad traffic. Transmission is blocked until fill is loaded. It has no jamming resistance.
* **PT — Plain Text**: the fallback mode. It uses 0.5× battery, but anyone can read it, jamming affects it one tier harder, and it has the largest DF signature.

The radio also includes a receive-only monitor mode, a scan mode that automatically changes the active preset, a squelch setting from 0–5 with a default of 3, and callsign entry with faction-specific suggestions. Its server-side net log stores the latest 50 entries and persists when the panel is closed or the radio is powered off. The client can filter the log by net or search sender and message text.

A **TX power** dial with LO, MED, and HI settings trades range for battery use and DF exposure. LO provides 0.6× range at half battery cost and half DF signature. HI provides 1.5× range at twice the battery cost and twice the DF signature.

The radio uses a swappable battery, with separate standby drain and a larger cost for each transmission or radio check. It powers itself off when the battery is empty. The antenna slot accepts either the stock **whip**, which keeps its range while moving, or the **mast**, which roughly doubles range while stationary but falls to around half the stock range while moving. That makes the mast useful at fixed positions without making it the best option in every situation.

The **RADIO CHECK** action sends a standard check over the active net and privately reports which stations are in clear or degraded range. When used inside a jam field, it also reports a compass bearing to the strongest emitter, giving players a way to locate jammers.

### Retrans stations

The radio can also relay while planted. An RTO can deploy it as an unattended retrans station, where it anchors its loaded faction nets, continues logging traffic, counts as stationary for the mast antenna, and drains its battery until recovered.

Anyone can pack up a retrans station, including the opposing faction. Doing so transfers the keyed radio and any loaded fill card to the person who recovers it. That is the intended capture risk.

### COMSEC

COMSEC uses physical fill cards loaded into a dedicated slot. Faction nets are encrypted end to end: an interceptor tuned to the correct frequency still hears headset traffic as static unless they have a matching, current-generation fill card.

The AN/PRC also needs valid fill to participate in that encryption. Attempting to transmit on a faction net without it warns the operator and sends the message in the clear. **ZEROIZE** ejects the card. **DESTROY** deletes it when recovery is not practical.

Captured enemy fill cards can be loaded to decode hostile secured traffic, including CT and the FH hopset. The countermeasure is the **RECRYPTO** order. With a current card from the operator's own faction loaded, RECRYPTO advances the faction-wide generation and invalidates every older card, including captured copies. Invalid cards remain identifiable as stale but no longer secure or decode traffic.

A captured enemy card cannot be used to trigger a recrypto for another faction. The order also does not create replacement cards. Newly spawned cards are stamped with the current generation, so normal resupply remains the way to distribute valid fill after a changeover.

### Direction finding

Traffic with a DF signature has a chance to reveal the sender's approximate location as a temporary contact on the opposing faction's tactical map. PT has the strongest signature, unsecured faction traffic is close behind, secured FH has a small signature, and secured SC or CT has none.

The detection chance scales with TX power, increases near active enemy jamming, and accumulates when a sender repeatedly transmits in the clear from the same position. Going quiet or moving away resets the accumulated exposure. The intended behavior is straightforward: transmit, then displace.

### Headsets and jamming

Every headset gains one tunable direct frequency from `30.000` to `87.999 MHz`, transmitted with `:x`. Direct-frequency traffic works peer to peer without towers, encryption keys, or relays. It is intentionally insecure: anyone can tune to the same frequency. An AN/PRC configured to that frequency can bridge the traffic over a longer distance.

This includes a headset tuning window, a reset-to-default verb, and a laminated frequency-assignment card item.

Radio jammers now interact with the full system. Standard radios are still hard-blocked when transmitting from inside a jam field. AN/PRC traffic is degraded instead, using mode-dependent garbling. A receiver near a jammer can add further corruption, degraded messages always lose at least one word, and direct frequencies are also affected.

### The handset

Nearby players can take the corded handset from a worn AN/PRC. The handset is a physical item using the existing RMC radio-telephone sprite and is connected to the radio by a 2.5-tile cord rendered on the client. This lets a commander or another nearby player speak through the RTO's pack without carrying a second radio.

While holding the handset, normal spoken messages are transmitted over the pack's active net. Whispers remain local, following the same behavior as the RMC rotary phones, and no `:r` prefix is required. The holder also hears the active net through the handset, excluding channels already provided by their own headset to avoid duplicate messages.

Dropping or throwing the handset returns it to the radio. Taking it from another player's hand transfers the call. Handset traffic uses the holder's callsign when available, otherwise it falls back to the station callsign.

### Comms arrays and masts

The mappable `AU14CommsArray` and `AU14CommsMast` structures provide fixed relay coverage. The array is powered and intended for whole-site coverage; the passive mast covers a smaller command-post area. Both can be damaged, destroyed, and repaired, with damaged sprite states at higher damage levels. This makes communications infrastructure a meaningful objective rather than background map decoration.

`RadioChannelPrototype` gains an `AnchorGated` flag. A combat net marked with this flag has **no coverage** unless an array, mast, worn AN/PRC, or planted retrans station is in range. Channels without the flag retain their existing behavior. `AnchorGated` does nothing while the CVar is disabled.

### Callsigns

GOVFOR, OPFOR, and CLF members receive a callsign based on job and squad when they spawn. The standard suffixes are 6 for element leader, 5 for second in command, 7 for senior NCO, ROMEO for RTO, OPS for staff, and 1–N for everyone else. These are prefixed by the squad word or the faction command word: HAVOC, VICTOR, or CELL.

On faction radio traffic, the callsign replaces the speaker's name and job prefix, following FH84-style net anonymity. During an AN/PRC transmission, the radio's station callsign takes precedence over the operator's personal callsign. Enabling the CVar in the middle of a round assigns callsigns to players who spawned while it was disabled.

Overwatch laptops receive a **comms net directory console** so command can resolve the anonymous callsigns within its own faction. Entries are grouped by element, and authorized users can reassign suffixes. Access is limited to AN/PRC operator roles.

### Role doctrine

Combat nets marked `AnchorGated` require relay coverage. The AN/PRC is restricted to RTO and command roles, and squad leaders no longer receive a direct command-channel key. SL-to-command traffic is expected to pass through the RTO or the handset.

This is the gameplay change that gives the RTO a necessary operational role rather than making the radio an optional accessory.

### Content

GOVFOR, OPFOR, and CLF each receive faction-specific AN/PRC variants, fill cards, and antennas. Trained roles carry their faction's fill card. RTOs spawn with a pre-keyed radio, a spare fill card, and a laminated frequency-assignment card.

Replacement radios, batteries, fill cards, and antennas are available through requisitions or ASRS for GOVFOR and OPFOR, and through faction vendors for CLF. A dedicated resupply crate is also included.

Five guidebook pages under a Communications Reference index cover the AN/PRC-117G, field headsets, voice procedure, and a numbers-only quick reference containing prefixes, ranges, and mode, power, and squelch tables.

## Why / Balance

This gives the RTO something important to manage throughout a round. Squads benefit from staying within relay range, losing an RTO can disrupt an entire element, and capturing a radio, retrans station, or fill card provides useful intelligence. The same mechanics could also support dedicated intelligence objectives later.

The COMSEC loop—fill, intercept, zeroize, capture, and recrypto—adds faction-level signals gameplay without changing the basic headset workflow for regular soldiers.

The controls are built around tradeoffs rather than a single optimal configuration:

* Modes: FH trades battery and a small DF signature for jamming resistance. SC is efficient and quiet while secured but has no protection from jamming. CT is difficult to detect or intercept but only works between filled AN/PRCs. PT is cheap and reliable as a fallback, but everyone can read it.
* TX power: additional range costs both battery and a larger DF signature.
* Antennas: the mast provides much better stationary coverage but performs worse than the whip while moving, keeping it useful for command posts and retrans sites rather than patrols.
* DF accumulation discourages repeated in-the-clear transmissions from one position and rewards basic signals discipline.
* Callsign-only faction traffic keeps intercepted messages useful without exposing player identities for free. The directory console preserves that ambiguity for the enemy while still giving friendly command staff the information they need.

The scope is deliberately limited. A relay only carries the operator faction's own nets. Colony and third-party channels continue to use their existing tower-dependent routing and are not affected just because an RTO tunes to them. Those groups should see no behavior change unless a jammer is involved, which affects all radio users consistently.

## Review notes

* The branch is split into **13 commits, one per subsystem**, in dependency order: shared declarations → crypto → garble/range → tunable headsets → manpack core → transmit/DF → handset/arrays/doctrine → physical handset item → server BUI → client window → callsigns → directory console → guidebook/vendors. Reviewing the combined "Files changed" view is also fine.
* Each commit builds independently and was run through the YAML linter on its own.
* For a server trial, enable `au14.new_comms_system`. All testing notes below assume the CVar is enabled.

## Please test this yourself

Please do not rely on code review alone for this one. It changes several layers of radio behavior, and the most useful feedback will come from trying unusual setups in-game.

Good things to try:

* [ ] Use two or more RTOs on the same faction net.
* [ ] Put different faction RTOs close to each other and check what does and does not relay.
* [ ] Try using the AN/PRC while moving between grids, maps, dropships, shuttles, or teleports.
* [ ] Kill, crit, cuff, strip, disconnect, or move the RTO and check whether relay behavior stops correctly.
* [ ] Put the radio in containers, drop it, turn it off, delete it, or move it away from the operator.
* [ ] Try friendly fill, enemy fill, no fill, zeroized fill, and stale (recrypto'd) fill.
* [ ] Order a recrypto and confirm captured cards stop decoding while fresh supply arrives current.
* [ ] Put jammers near the sender, receiver, relay, or all of them at once.
* [ ] Transmit repeatedly from one spot in PT near jamming and watch the enemy tactical map for DF blips; then displace and confirm the accumulation resets.
* [ ] Plant a retrans station, walk out of worn-relay range, and confirm it carries the net; pack it up as the enemy faction.
* [ ] Swap whip/mast antennas while moving and standing still, at each TX power, and confirm the range changes (radio check helps here).
* [ ] Drain a battery to zero mid-conversation and hot-swap it.
* [ ] Try direct-frequency headset comms between hostile groups using the same frequency.
* [ ] Take the handset, walk the cord to its limit, talk and whisper while holding it, drop it, throw it, and grab it out of someone else's hand mid-sentence.
* [ ] Confirm the handset holder hears the net without double-hearing channels their own headset already covers.
* [ ] Destroy and repair a comms array/mast and confirm anchor-gated nets die and come back.
* [ ] Check callsign assignment on roundstart, latejoin, squad transfer, and mid-round CVar enable; confirm names never leak on faction nets.
* [ ] Use the directory console to look up and reassign suffixes; try it as an unauthorized role.
* [ ] Check implants, accessory earpieces, headset encryption keys, muted channels, ghosts, and admin observing.
* [ ] Fill the net log past 50 entries with several nets loaded, then filter by net and by text with the panel open while traffic keeps arriving.
* [ ] Spam several relays, several jammers, and many different direct frequencies to see if anything breaks.

Please add any missing edge cases directly to the PR. I would rather test too many odd interactions than miss an obvious exploit.

## Ingame testing performed

The following has already been tested ingame:

* [x] AN/PRC-117G can be equipped on the back slot.
* [x] Only trained operators can operate the AN/PRC-117G.
* [x] Radio presets can be configured from faction nets and through raw frequency entry.
* [x] `:r` transmits on the selected active AN/PRC net.
* [x] Friendly headset traffic is relayed by a worn, powered AN/PRC; relay coverage works without a communications tower within the intended relay range, and degrades outside full-signal range.
* [x] Tower-gating exemptions work for relay-covered traffic.
* [x] The relay only carries the operator faction's own nets; tuning enemy, colony, or third-party frequencies does not reroute those channels for uninvolved players.
* [x] Direct headset frequency tuning works within `30.000–87.999 MHz`; `:x` transmits peer-to-peer without towers; AN/PRC bridging works for shared frequencies.
* [x] Matching COMSEC fill decodes secured faction-net traffic; missing or wrong fill hears static; captured enemy cards decode hostile secured traffic; intercepts appear as red intercept text.
* [x] ZEROIZE ejects the card; DESTROY deletes it.
* [x] RECRYPTO supersedes older cards faction-wide; stale cards stop securing and identify as STALE; enemy cards cannot trigger a recrypto for your faction.
* [x] Jammers hard-block standard radios from inside a jam field but degrade AN/PRC transmissions instead; FH/PT apply their intended resistance/susceptibility; direct frequencies are jammable.
* [x] TX power scales relay range, battery cost, and DF exposure at each setting.
* [x] DF blips appear on the opposing faction's tactical map for in-the-clear/PT traffic near jamming; accumulation resets on displacement or going quiet.
* [x] Retrans plant/pack-up flow, including enemy capture and mast-on-retrans range.
* [x] Handset take/hang-up/cord-break/drop/throw/strip flow, all-speech transmit with whisper staying local, holder hearing without duplicates, cord overlay.
* [x] Callsign assignment and radio name masking on faction nets, station callsign override mid-AN/PRC transmission.
* [x] GOVFOR and CLF variants spawn with their expected faction equipment.
* [x] Guidebook pages open and display correctly.

Still to tick off:

* [ ] Radio check coverage report, including the jammer bearing while inside a jam field.
* [ ] Battery drain on standby vs. per transmission, auto power-off, and hot-swap.
* [ ] Whip vs. mast antenna ranges while moving and stationary.
* [ ] Net log filters (per-net dropdown, text search, CLR) with live traffic arriving.
* [ ] Directory console suffix reassignment across squad transfers.
* [ ] Comms array/mast damage states, destruction, and repair, with anchor-gated nets dropping and recovering.

## Merge warning

> [!WARNING]
> Merging is safe at any time because the CVar is disabled by default, and the current radio behavior remains unchanged while it is off. However, **do not enable `au14.new_comms_system` until maps have comms coverage placed**. Faction main bases and the ship need an `AU14CommsArray`; command posts need an `AU14CommsMast`. Anchor-gated combat nets will not work without one of those structures, a worn manpack, or a planted retrans station. Mapping can be merged separately, but it must be complete before the CVar is enabled.
>
> This change touches radio routing, tower gating, jamming, faction encryption, the tactical map, item slots, headset behavior, and intercept logic. It needs broad in-game testing. The proposed server trial is the right way to validate it.

## Edge cases reviewers should think about

Please reply to this PR with any abuse cases, edge cases, or unusual interactions you can think of, especially around:

* [ ] Multiple RTOs relaying the same net on the same map, and different-faction RTOs near each other.
* [ ] AN/PRC relay behavior across grids, maps, shuttles, dropships, and FTL/teleport movement.
* [ ] Radio behavior when the RTO dies, disconnects, is crit, is cuffed, is stripped, or drops the radio.
* [ ] Relay behavior when the AN/PRC is powered off, unworn, deleted, gibbed, EMPed, or moved into containers.
* [ ] Captured radios with friendly fill, enemy fill, no fill, zeroized fill, or stale fill.
* [ ] Recrypto abuse: ordering changeovers with captured or stale cards, recrypto spam, and cards spawned mid-changeover.
* [ ] Retrans stations: planting in odd places, capture griefing, planted radios on moving grids, and mast range while planted.
* [ ] Handset abuse: cord range on moving grids, handing the handset through walls/windows, holder identity leaks, hearing duplication with exotic headsets.
* [ ] Callsigns: name leaks through emotes/whisper/examine, suffix collisions on mass latejoin, squad transfer races, mid-round CVar enable.
* [ ] DF blips: correct faction targeting, blip cleanup after expiry, and interaction with tactical map snapshots for non-live viewers.
* [ ] Ways to cheese the mast antenna's stationary check (stutter-stepping, being pulled or carried, standing on moving grids/conveyors).
* [ ] Jamming behavior when the transmitter, receiver, relay, or all three are inside different jam fields.
* [ ] Direct-frequency abuse cases, especially shared frequencies between hostile groups.
* [ ] Interaction with implants, accessory earpieces, headset encryption keys, channel mutes, and ghost/admin observing.
* [ ] UI desyncs between the radio panel, headset tuning window, item slots, net log, net log filters, and the directory console.
* [ ] Roundstart loadouts, latejoin loadouts, role changes, respawns, and admin-spawned equipment.
* [ ] Performance with many players, many direct frequencies, several jammers, and multiple active relays.
* [ ] System ordering: `ANPRCRangeSystem` relies on `after:` constraints against `RMCTelephoneSystem` / `JammerSystem`. If upstream ever renames those systems, this fails loudly at server startup rather than silently.
* [ ] Engine subscription constraints: only one system may subscribe a given `(component, event)` pair. If upstream ever subscribes `<HeadsetComponent, MapInitEvent>` itself, the server should fail loudly at startup with a `Duplicate Subscriptions` exception instead of silently changing behavior.

## Technical details

System layout:

* `ANPRCRadioSystem` (split into core / `.Transmit.cs` / `.Handset.cs` / `.Bui.cs` partials): handles equip and unequip, channel grants, relay-anchor lifecycle, battery and antenna slots, presets, transmit and receive paths, interception, DF exposure, retrans, handset behavior, and the net log.
* `ANPRCCryptoSystem`: handles the fill-card item slot, fill generations, and recrypto changeovers. RECRYPTO is gated on an `OperatorFaction` match.
* `ANPRCGarbleSystem`: applies jamming, range, COMSEC garble tiers, and squelch. Jam intensity is cached once per tick.
* `ANPRCRangeSystem`: handles per-map relay range, antenna and TX-power effective ranges, coverage classification, and `TryGetNearestJammerDirection` for the radio-check bearing.
* `TunableFrequencySystem`: routes direct-frequency traffic and AN/PRC bridging while mirroring `RadioSystem` bookkeeping, including admin logs, replay recording, and ghost hearing with the follow button.
* `AU14CallsignSystem` (+ `.Console.cs`): assigns callsigns, preserves pinned suffixes across squad transfers, masks radio names, and backs the directory console. Suffixes are unique per element.
* `TacticalMapSystem.AU14.cs`: creates DF contact blips in both the live and snapshot dictionaries because non-live tactical-map viewers only render the snapshot.

Notes:

* `ANPRCRangeSystem` subscribes to `RadioSendAttemptEvent` / `RadioReceiveAttemptEvent` **after** `RMCTelephoneSystem` and `JammerSystem`, then overrides their verdict for relay-covered traffic and AN/PRC `:r` transmissions. Conditions that must remain cancelled, including muted channels through `RMCRadioFilter` and jammed non-ANPRC senders, are checked again before a cancellation is removed.
* `ANPRCGarbleSystem` subscribes to `RadioReceiveEvent` on `TunableHeadsetComponent` **before** `HeadsetSystem`. The by-ref event passes the garbled message into the normal relay-to-chat path.
* `TunableFrequencySystem` adds `TunableHeadsetComponent` and registers its BUI on each headset during map initialization through `SharedUserInterfaceSystem.SetUi()`. This avoids modifying the upstream `CMHeadset` prototype.
* **Engine constraint:** only one system may subscribe to a given `(component, event)` pair. The `_AU14` systems therefore use `TunableHeadsetComponent` for receive garbling and `<HeadsetComponent, MapInitEvent>` for component injection. If upstream later adds its own `<HeadsetComponent, MapInitEvent>` subscription, the server will likely fail at startup with a `Duplicate Subscriptions` exception. The fix would be to retarget our subscription.
* The radio item owns its `ActiveRadioComponent`, like a headset, rather than granting channels directly to the wearer. This avoids conflicts with accessory earpieces or radio implants that add `IntrinsicRadioReceiverComponent` to the player.
* Callsign masking uses a transient per-tick flag around `SendRadioMessage`. `TransformSpeakerNameEvent` is subscribed with `after: [typeof(RankSystem)]`, and `JobPrefixComponent` is removed and restored around the send. The handset speaker-name mask is unconditional so a timing race cannot expose the holder's rank or real name over the radio.
* The handset is a physical entity based on the `SharedRMCTelephoneSystem` rotary-phone pattern. It is spawned into a `ContainerSlot` on the pack during MapInit, picked up with `TryPickupAnyHand`, and returned to the pack on hang-up, cord break, or drop. All-spoken-message transmission runs from the `EntitySpokeEvent` handler with quiet validation, so an invalid pack state falls back to local speech rather than producing an error. The holder mirrors the pack's granted receive channels through `ActiveRadio` and an intrinsic receiver synchronized every tick, skipping channels already covered by the holder's headset. `ANPRCHandsetOverlay` renders the cord client-side and mirrors `TelephoneOverlay`.
* Comms arrays and masts use `AU14CommsStructureVisuals` for damage-state sprites. The broken-tower state begins at 250 damage, and destruction occurs at 450 of 500 damage. Both structures can be repaired.
* Mode and TX-power multipliers for battery use, range, and DF live in shared `RadioModeExtensions` / `RadioTxPowerExtensions`, keeping the client UI and server calculations on the same values.
* Antenna ranges are copied from the inserted antenna item onto the relay anchor. A stationary-only antenna used by a moving wearer falls back to stock range multiplied by its `MovingRangeMultiplier`, which is 0.5 for the mast. Planted retrans stations always count as stationary.
* The net log is stored server-side, capped at 50 entries, and persists while the UI is closed or the radio is powered off. The net dropdown and text search only filter the client view and do not modify the stored log.
* `RadioChannelPrototype` gains an `AnchorGated` boolean. The master branch already carries fork-specific fields such as `Tower`, `Planet`, and `Faction`, so this follows the existing repository pattern.
* Added BUIs and XAML windows for the radio panel, headset frequency tuning, and the callsign directory console.
* Added prototypes for faction-specific AN/PRC variants, batteries, whip and mast antennas, fill cards, the frequency-assignment card, the handset, comms array and mast structures, faction net channels, a resupply crate, requisitions catalog and vendor entries, and sentinel radio channels used by `:r` and `:x` routing.
* Added guidebook XML for the Communications Reference index, AN/PRC-117G, Field Headsets, Voice Procedure, and Comms Quick Reference, along with FTL locale files.
* Updated role prototypes for AN/PRC operator access, channel assignments, and removal of the squad-leader command key.

## Media

image
image
image
image
image
image
image
image
image
image

## Requirements

* [x] I have read and am following the [[Pull Request and Changelog Guidelines](https://docs.spacestation14.com/en/general-development/codebase-info/pull-request-guidelines.html)](https://docs.spacestation14.com/en/general-development/codebase-info/pull-request-guidelines.html).
* [x] I have added media to this PR or it does not require an ingame showcase.
* [x] I have included a **brief** detail of the major changes for this PR in the changelog below.
* [x] By submitting this code and/or assets, I confirm that I either own them or have provided the correct necessary licenses to use and distribute them. I agree to be fully responsible for any legal claims or issues arising from the use of these materials.

**Changelog**
:cl:

- add: Added the AN/PRC-117G manpack radio for Radio Telephone Operators and command roles, with up to four programmable preset nets, operator callsigns, and `:r` transmission.
- add: Added relay-based ground comms. Headset traffic on nets loaded in a worn AN/PRC works without a comms tower within 30 tiles at full signal and 45 tiles with degraded signal.
- add: Added AN/PRC operating modes: Frequency Hopping, Single Channel, Cipher Text, and Plain Text. Each mode trades battery cost, jamming resistance, security, and direction-finding signature, so no single mode dominates.
- add: Added a transmit power dial (LO/MED/HI) to the AN/PRC, trading relay range against battery drain and direction-finding exposure.
- add: Added physical COMSEC fill cards for GOVFOR, OPFOR, and CLF. Secured faction-net traffic is pure static to interceptors without the matching fill, and captured enemy cards can be loaded to decode hostile traffic.
- add: Added the RECRYPTO order, a faction-wide COMSEC changeover that supersedes all previously issued fill cards, including captured ones.
- add: Added direction finding. Transmitting in the clear, in PT mode, or near enemy jamming can reveal the sender's position as a temporary contact blip on the opposing faction's tactical map, with repeated transmissions from the same spot raising the odds.
- add: Added retrans stations. A planted AN/PRC relays its faction nets unattended, logs traffic, and drains its battery. Anyone, including the enemy, can pack it up and capture it.
- add: Added signal interception, monitor mode, scan mode, a squelch noise gate, and a filterable 50-entry net log to the AN/PRC.
- add: Added a battery slot and an antenna slot to the AN/PRC. The whip antenna holds relay range on the move, while the mast roughly doubles range when stationary but halves it while moving.
- add: Added a radio check that reports which stations are in clear or degraded range and, inside jamming, gives a bearing to the strongest emitter.
- add: Added a tunable direct frequency from 30.000 to 87.999 MHz, transmitted with `:x`, to all headsets. It works peer-to-peer without towers or keys and can be bridged by an AN/PRC.
- add: Added faction-wide automatic radio callsigns that replace names on faction nets, plus a callsign directory console on overwatch laptops.
- add: Added comms arrays and masts that anchor faction nets over an area. Combat nets require relay coverage from an AN/PRC or array.
- add: Added AN/PRC variants, fill cards, and antennas for GOVFOR, OPFOR, and CLF, with resupply through requisitions and faction vendors.
- add: Added guidebook pages for the AN/PRC-117G, field headsets, voice procedure, and a comms quick reference.
- add: Added role tags on the net for squad leaders and fireteam leaders, shown as `(SL) ALPHA 6` and `(FTL) ALPHA 2-1`.
- add: Added fireteam-based callsigns. A marine's fireteam forms the first half of their numeric suffix, so `ALPHA 2-3` is fireteam two, member three. Callsigns update automatically on reassignment.
- add: Added aircrew callsigns. Pilots answer to `PAPA`, and dropship crew chiefs answer to `CHIEF` within the command element.
- add: Added task groups to the callsign directory console. Net control can create custom-word groups such as `REAPER`, move personnel in and out, and disband them, with callsigns updating automatically.
- add: Added role sections to the callsign directory. Aircrew, military police, medical, and intelligence are listed separately from the command element and only appear when staffed.
- add: Added a buildable field antenna mast, raised from ten metal sheets by engineering-trained personnel, that relays both factions' nets.
- add: Added hull-mounted vehicle radio sets to dropships and vehicle interiors, including APCs, Humvees, tanks, and the Blackfoot, keeping mounted troops on the net at short range.
- add: Added your callsign to your headset's examine text.
- add: Added per-operation frequency plans: faction net frequencies are randomized every round, and frequency cards print the current assignments — captured cards are now real intel.
- add: Support synthetics are now radio trained and can operate the AN/PRC.
- tweak: Radio jammers now degrade AN/PRC transmissions with mode-dependent garble instead of blocking them. They jam both endpoints and also affect direct frequencies.
- tweak: Squad leaders no longer have a direct command-channel key. SL-to-command traffic now flows through the RTO or the AN/PRC handset.
- tweak: RECRYPTO now requires command COMSEC authority: Commander, Adjutant, Advisor, or Cell Leader.
- tweak: MP nets now use the relay network like every other faction net instead of depending on the legacy comms tower.
- tweak: Ambassadors no longer receive callsigns or appear in the net directory.


maertayn:
Who told you splitting one interconnected system into 12 PRs is a good idea?

rokudara-sen:
> Who told you splitting one interconnected system into 12 PRs is a good idea?

The original #1306 had grown to 8k lines and anything this big is awful to review, so I split it along the dependencies

Each part builds independently, and the full feature remains disabled behind the CVar, which should make partial merges safe

I’m happy to do it in whichever format is easiest for the you guys to review, since I come from a background where doing a feature in small packages is the norm, I thought it would be the best course of action. So, we could do:

1. The current stack of 12 small PRs
2. A single PR with 12 commits
3. A middle ground

There’s no meaningful rework required on my side anymore. Whoever plans to merge it can choose the format

maertayn:
If each PR works standalone, then yeah perfect, but I fear you're slitting 8k lines across 12 different PRs which makes review even worse than just 1 monster

rokudara-sen:
> If each PR works standalone, then yeah perfect, but I fear you're slitting 8k lines across 12 different PRs which makes review even worse than just 1 monster

Yeah, no, each part is self-contained. I compiled eached, ran it through the YAML linter and all. The CVar is off anyhow, meaning all changes will be disabled regardless. 

On the total review load, the parts are disjoint and together make up the same 8k line feature. Split doesn't reduce anything, it just keeps each review focused on each subsystem

Again, if you want me to I can reopen the single PR and have like 12 or 13 commits for review there

maertayn:
One PR is best that way there's no juggling.

Commits are nice, not needed, usually just look at all files changed regardless of commits (it shows the end result and not the rewrites someone took to get there)

For big ones, checkout the PR and read it in nvim as I have tools written for it that makes life easier.

After a review, PR this size. Probably will run a test merge first and try it out on the server for a few days. 
That surfaces a lot of issues you and I might have overlooked, and it reduces some burden as I don't have to intricately understand why you wrote each system the way you did.

maertayn:
I'm reading that all maps require 'AU14CommsArray' or 'AU14CommsMast'.

To circumvent that requirement, can these be spawned in mid-round, and the Cvar toggled to enable for a test merge run of this?

rokudara-sen:
> I'm reading that all maps require 'AU14CommsArray' or 'AU14CommsMast'.
> 
> To circumvent that requirement, can these be spawned in mid-round, and the Cvar toggled to enable for a test merge run of this?

Yes, that should work.

Coverage anchors are checked whenever a message is sent. Nothing is registered only at round start. You can admin-spawn the required entities mid-round and then enable the system with the Cvar

For a test merge, place an `AU14CommsMastGovfor`. These masts require no power and provide full coverage for all faction combat nets within 75 tiles, with degraded coverage extending to 100 tiles

The Bush will also need an `AU14CommsArray` or comms mast, otherwise it will have no radio coverage. On the Bush specifically, use the array rather than a mast. Ship decks are separate z-levels, and coverage is now z-aware, meaning arrays cover their entire connected z-network and masts, worn radios, and retransmission stations only reach one level above and below

As a result, one array placed anywhere on the Bush will cover all decks, while a mast would only cover its own deck and the decks directly above and below it

Callsigns are assigned retroactively to players who have already spawned. Toggling the CVar off will also revert everything

One note:
I removed the command encryption key from the SL headset, so that change remains regardless of the CVar

maertayn:
"For GOVFOR, those roles are the RTO, Commander, Officer Adjutant, and Advisor; for CLF, the Cell Leader can operate it. This redundancy is intentional: if the RTO goes down, another trained role can pick up the keyed radio and keep the element connected."

What redundancy? If the RTO goes down, comms are done for, no one else is skilled to use it in the field..?

rokudara-sen:
> "For GOVFOR, those roles are the RTO, Commander, Officer Adjutant, and Advisor; for CLF, the Cell Leader can operate it. This redundancy is intentional: if the RTO goes down, another trained role can pick up the keyed radio and keep the element connected."
> 
> What redundancy? If the RTO goes down, comms are done for, no one else is skilled to use it in the field..?

Troy told me that there can be two RTOs per squad

rokudara-sen:
Since the last comment, I rebased the branch onto current master and force-pushed it. All commits are intact, with no merge commit

The RTO comms kit has been moved from the old Radio Telephone Operator slot to the new Radio Traffic Operator. The new role no longer spawns with a PRC-55. I also added the missing OPFOR Traffic Operator, with matching gear, ID, and three slots

Mid-round CLF recruits now receive callsigns instead of using their civilian names. The GOVFOR Advisor no longer gets an OPFOR COMSEC fill, and the OPFOR Advisor now gets the correct one

Squad leaders and fireteam leaders are now tagged on the net, for example, “(SL) ALPHA 6” and “(FTL) ALPHA 2-1.” Fireteam assignments also control the numeric suffix, and callsigns update when players are reassigned

Pilots now use PAPA, DCCs use CHIEF, and the directory console can create, disband, and manage custom task groups such as REAPER

RECRYPTO is now limited to command roles. Engineers level 2 and above can build a field antenna mast from 10 metal, and it relays both factions’ nets

All dropships and vehicle interiors now have a small, always-on relay, including APCs, Humvees, tanks, SPP vans, the Blackfoot, and the ARV. Vehicle crews were previously cut off from the network

Further playtesting also fixed ghost and admin ghost reception, mast rotation, and the AN/PRC panel not refreshing when the pack was equipped or removed

Examining your headset now shows your current callsign

PR #1573, opened 2026-07-11
Add 2 hour limit to Insurgency gamemode

deepdarkdepths:
## About the PR
There was a round last night where CLF did nothing for so long that almost all of govfor just disconnected from boredom. Insurgency rounds should not be lasting 3+ hours with nothing happening. This aims to both force CLF to get their shit together and do something, and make people more willing to participate in the round knowing that something will happen before the 2 hours is up. 2 hours is also plenty of time to think of and carry out a plan.
In addition, some people are employed and do not have the time to wait 3 hours or more for the round to end. Me personally, I would rather not wait 3 hours just to play the gamemodes I like.

There is probably a better way of going about this but right now this solves the problem of any remaining CLF sitting around doing nothing.

## Technical details
TimeSpan is hardcoded but it shouldn't matter much.

## Requirements

**Changelog**

:cl:
- add: Insurgency now ends with a neutral outcome at 2 hours.

### DIFF
--- a/Content.Server/_CMU14/Threats/Rules/InsurgencyEndRuleSystem.cs
+++ b/Content.Server/_CMU14/Threats/Rules/InsurgencyEndRuleSystem.cs
@@ -8,6 +8,8 @@ public sealed class InsurgencyEndRuleSystem : GameRuleSystem<InsurgencyRuleCompo
 {
     [Dependency] private readonly GameTicker _gameTicker = default!;
 
+    private const string DefaultWinMsg = "Neutral outcome: The CLF have failed to gain a significant foothold in the colony during their time there. The cell and the military fall into a stalemate, both sides silently operating in the colony for days to come.";
+
     private static readonly TimeSpan RoundTimeLimit = TimeSpan.FromHours(2);
 
     public override void Update(float frameTime)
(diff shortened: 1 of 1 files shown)


deepdarkdepths:
My cl is fine bro
Bot tweaking

rokudara-sen:
Could we avoid ending the round immediately when the two-hour limit is reached?

Instead, consider announcing that the round will end after a short delay, similar to the upstream evacuation shuttle countdown. During that window, we could open a vote allowing players to either continue the round or proceed with the stalemate ending

If the vote passes to continue, the round could remain active and the same vote could be triggered again after another 30-60 minutes. This could repeat until the players eventually vote to end the round or it ending naturally

That would make the transition less abrupt and give the players some control over whether the round has genuinely run its course

maertayn:
I agree that an abrupt irreversible end after 120 minutes does not seem feasible.

deepdarkdepths:
> Could we avoid ending the round immediately when the two-hour limit is reached?
> 
> Instead, consider announcing that the round will end after a short delay, similar to the upstream evacuation shuttle countdown. During that window, we could open a vote allowing players to either continue the round or proceed with the stalemate ending
> 
> If the vote passes to continue, the round could remain active and the same vote could be triggered again after another 30-60 minutes. This could repeat until the players eventually vote to end the round or it ending naturally
> 
> That would make the transition less abrupt and give the players some control over whether the round has genuinely run its course

I discussed this with a few people before and there were some concerns about the round being continued by vote endlessly by either govfor or clf, completely nullifying the point of a vote existing if it doesn't encourage ending the round. I don't really like the concept unless we implement really low requirements for a round to end, such as only 40% of votes compared to player population needed to end the round.

PR #1574, opened 2026-07-11
Chat revamping and markup larp,

NileDeveloper:
Able to bold italic and bold italic individual words parts of sentences

Word* makes it bold, Word/ makes it italic

sentence **sentence** sentence makes the sentence bold

Same thing with italic

Sentence //sentence// sentence,

to make something both bold AND italic end it with ***

word*** word

Example sentence

HELP* me/ get ** *out now* ** because "this'* place is //collapsing// fast***

made and added cozzete italic, at 20 degrees, may look better at 24 degrees for those that want to make changes later on,

and i honestly recommend that the normal chat font italic to be changed to be a bit sharper then it is right now since it is EXTREMELY subtle, and i honestly just wanted this to be out there and had zero energy to fix that,

Revamped basic chat for radio chatter to work the same as rune chat

at first radio chatter would show whisper and radio chatter at the same time 2 bubbles above the characters head, now its just the whisper above the character head similar to rune,

uuh yea i think thats it?

it doesnt work with command announcement works with radio and speech, works with xenos as well, 

tested it a couple of times as far as i know there should be no apparent bugs atleast,

## Requirements

- [ ] I have read and am following the [Pull Request and Changelog Guidelines](https://docs.spacestation14.com/en/general-development/codebase-info/pull-request-guidelines.html).
- [ ] I have added media to this PR or it does not require an ingame showcase.
- [ ] I have included a **brief** detail of the major changes for this PR in the changelog below.
- [ ] By submitting this code and/or assets, I confirm that I either own them or have provided the correct necessary licenses to use and distribute them. I agree to be fully responsible for any legal claims or issues arising from the use of these materials.

**Changelog**

:cl:
- add: individual boldening of words and sentences
- remove: double chat bubbles using normal speech instead of rune and making it similar to rune with one single speech bubble whisper with radio chatter
- add: individual italic of words and sentences
-add: individual italic  AND boldening of words and sentences
-revamped runechat so it works with individual boldening and italics instead of how it was before -all of it bold or none is bold-


maertayn:
What if you had a question and made a typo like this/
Your question becomes italic?

NileDeveloper:
> What if you had a question and made a typo like this/ Your question becomes italic?

between makingi t so you have to go to use the mark before and after the word or just end it with * i went ease with possibility of error ratherr then certainty with less ease

PR #1594, opened 2026-07-12
Fixed BUG-161

NileDeveloper:
Fixed bug-161 made emphasis work instead of appaearing /color due to markup,

:cl:

- fix: Bug-161 Runechat does not sanitize markup from input


maertayn:
Changelog bot scans for :cl: without it, your changes wont be found

Took the liberty of modifying your PR description for that

PR #1598, opened 2026-07-12
Replaces WICH for the Geneva Convention

IMGXKILLER:
Since the Geneva Convention is Canon in Aliens, and WICH received quite some negative backlash, I decided to replace WICH with a more sovlfull (And extended) Geneva convention

## About the PR
Replaces WICH with a Geneva Convention that's more Expanded and akind to real life.

## Why / Balance
It's soul

:cl:
- add: The Geneva Conventions of 12th August 2106
- remove: WICH


maertayn:
Pending head moderator or host review

maertayn:
Unless, did you change anything to it? If not, it should be good to go without review from them.

I did notice some formatting/markup strangeness

IMGXKILLER:
This is my first PR ever so no idea, it does remove WICH and add Geneva in it's place

maertayn:
If this is simply a rename, there is 't anything that requires them to reconsider.

SOP/rules/laws change the gameflow significantly their OK is needed so that staff isn't enforcing some obscure silly change that has no relevance in our community.

Unfortunately because the files have been renamed/moved I cannot see a diff of changes.

IMGXKILLER:
it is a full rewrite so that means itll probably need approval

if you want to compare the changes i think you could try looking up for the WICH pr

Crime481:
warfare interstellar charter of humanity sounds more futuristic and cool imo /shrug

IMGXKILLER:
perhaps, but aliens canonically has the geneva convention

twinklezzzz:
This looks good and very soulful

Crime481:
> perhaps, but aliens canonically has the geneva convention

ok then fair

PR #1600, opened 2026-07-13
"A Small YAML Change" and "INSFOR Update" PRs(unified)

wray-git:
## About the PR

Jerrie's Mod.

This adds three connected features to the construction system, all kept in `_AU14/` folders so they stay isolated from upstream:

1. **A new construction menu** (referenced from a funny game). Searchable, grouped by category, with a slide-out detail panel. The classic menu is still available and can be re-selected in Options.
2. **Saved Builds.** Select structures you built, save them to a file, and place them again later. Files can be shared with other players.
3. **Vertical building.** Dig down into a generated stone level, or build floors upward into open space on support beams. Over-mining a cavern or destroying a support beam causes the unsupported roof or floor to collapse.

It also adds a set of admin tools for adding items, tiles, and lathe recipes to the menu in-game without editing YAML by hand.

------------------
INSFOR:
- Converts (visually) CLF into INSFOR. All job names, codewise, remain the same.
- Introduces **INSFOR Editor**. The ability to create INSFOR factions, editing them completely and uniquely. From their names & descriptions, down to the contents of the vendors, including what is the entity(model/sprite/item) the vendor is.
- CLF leader now chooses which faction he wishes the INSFOR to be in the round. _**INSFOR Editor sets the GOVFOR whitelist for the faction(s)**_, meaning specific INSFOR factions can only be chosen if a specific GOVFOR faction is chosen for the round.
- CLF Leader's heavy cell kit reworked, to where they can now individually pick each item in the cell kit to be placed down.
- Introduces **"A Package"** item. An INSFOR Editor can add "A Package" which gets given to the jobs he selected(i.e INSFOR members). Contents of the package are set by the editor.
- Introduces **CLF Sapper**. A new INSFOR role that is capable of creating and using traps, alongside his own workbench which he can forceably screw attachments into weapons, even if they don't accept it.
- Introduces **Siphon Rig**. CLF Sapper's tool to steal colony funds.
- Each 1 construction skill point now gives 10% reduction in material consumption(metal, wood, plasteel), and increases building speed by 10%. Cades can now be ghost constructs once again, and everyone can now place foldable cades.
- INSFOR members are now able to use WS equipment they manage to capture. Though trying specific weapon(s) such as SHARP is ill-advisable...

## Why / Balance

Players can now save their buildings.
Admins(as of the PR, only Hosts) can now easily add any item into the construction menu. _**This is to encourage the team to add more constructables into the menu for more variety.**_ I want to see over 100 entities in the menu by end of the week.

---------------------
INSFOR:
- Converts (visually) CLF into INSFOR. All job names, codewise, remain the same.

- Introduces **INSFOR Editor**. The ability to create INSFOR factions, editing them completely and uniquely. From their names & descriptions, down to the contents of the vendors, including what is the entity(model/sprite/item) the vendor is.
> INSFOR Editor is currently locked behind the `Host` flag. There is an additional `InsforEditor` jobaddwhitelist available.
> This ensures that any INSFOR factions added are approved through the Host's discretion. 
- CLF leader now chooses which faction he wishes the INSFOR to be in the round. _**INSFOR Editor sets the GOVFOR whitelist for the faction(s)**_, meaning specific INSFOR factions can only be chosen if a specific GOVFOR faction is chosen for the round.
> The menu appears when he spawns in.
> Once again, available factions are based on what's the current GOVFOR faction in the round.
- Introduces **"A Package"** item. An INSFOR Editor can add "A Package" which gets given to the jobs he selected(i.e INSFOR members). Contents of the package are set by the editor.
> Allows for the INSFOR editor to give "loadouts" to INSFOR members without any complicated lobby-loadout codes, which wouldnt make much logical sense either.
> Also allows for the INSFOR Editor to create unique situations where this package can appear to *any* job, such as a letter sent to xyz individual.
- Introduces **CLF Sapper**. A new INSFOR role that is capable of creating and using traps, alongside his own workbench which he can forceably screw attachments into weapons, even if they don't accept it.
> CLF Sapper has the following equipment:
> - Tripwire trap: Able to attach up to 3 or 5(dont remember exact number) of any form of "activate-able" devices.
> - Audio trap: Sounds a whistle and alerts on the INSFOR comms when tripped.
> - Buried "activate-able device"
> - Snare trap: Flips the snared individual's POV by 180, traps for 60 seconds, unless freed with a knife(1s)
> - Spy Camera, Relay, Handheld Monitor
> - Siphon Rig: Siphons from colony budget computer, ATMs, and ASRS.
> - A switch: a unique attachment that... honestly don't craft it.
>
> The CLF Sapper will allow for much needed hard-counters against specific WSes(scout), and introduces a new avenue of income-gain without having to RR the colony admin.

- Introduces **Siphon Rig**. CLF Sapper's tool to steal colony funds.
> Allows for a less violent new source of income.

- Each 1 construction skill point now gives 10% reduction in material consumption(metal, wood, plasteel), and increases building speed by 10%. Cades can now be ghost constructs once again, and everyone can now place foldable cades.
> With the new Jerrie's Mod building PR, building is going to become much more common/accessible. This makes engineers truly distinct and sought after for their skills.
> I pray CTs actually build cadelines now with this.

- INSFOR members are now able to use WS equipment they manage to capture. Though trying specific weapon(s) such as SHARP is ill-advisable...
> It is already an achievement to take out a WS as INSFOR.
> The reason why WS equipment is locked is mainly to prevent other GOVFOR members from stealing WS equipment in Distress Signal. In Insurgency, this isn't necessary.
> Certain WS equipment such as SHARP still activates against INSFOR members.

## Technical details

All new code lives under `_AU14/` (Client/Server/Shared). Files that had to modify upstream code are listed at the end and are marked in-code.

### Z-level building: the values most likely to be tuned

The two numbers contributors will most often want to change:

**Support beam reach (cantilever span).**
Set per beam with `cantileverSpan` on the `StructuralSupport` component, in `Resources/Prototypes/_AU14/ZLevelBuilding/supports.yml`:
- Blue beam = `2` (also acts as the ground anchor for the stairs. **Dont touch**, it acts slightly different than other support beams)
- Brown beam = `3`
- Green beam = `5`
- Aqua beam = `8`

A larger value lets floors and structures extend further from the beam before they lose support and fall. Any plain built structure defaults to `2` (`StructuralSupportComponent.CantileverSpan`). To make beams reach further, raise the number.

**Cave-in range.**
`MaxRoofSpan` on the `ZBuildableMap` component (default `3`). Any dug-out tile more than this many tiles from the nearest solid rock or built pillar has an unstable roof and collapses after an 8 second warning. Placing a vertical pillar makes a wider span safe, the same way mine support pillars work. Raise `MaxRoofSpan` for more forgiving caverns, lower it for tighter ones.

**Other `ZBuildableMap` values** (all `[DataField]`, per map, in `ZBuildableMapComponent.cs`):
- `ChunkSize` (default 8): size of each lazily generated stone chunk below.
- `StoneFloorTiles`: tiles used for the generated underground floor.
- `StoneRockEntity`: the mineable rock scattered in generated chunks.
- `RockDebris` / `CollapseFog` / `CollapseRockProp` / `RumbleSound`: the cave-in effects (flung rock, fog, falling-rock prop, crash sound).

### Components

- **`StructuralSupport`**: marks anything that participates in the support graph. `cantileverSpan` is its reach, `isAnchor` makes it a permanent ground root, `isVerticalSupport` makes it a pillar/beam that relays its full span onward (chaining beams extends reach), `HideFromScanner` is used on staircase beams so the scanner does not highlight their tiles as buildable (they only support the stairs), and `Supported` is the runtime state.
- **`ZBuildableMap`**: the per-map on/off switch plus all generation and collapse values above.
- **`ZStair` / `ZStairBeamLink`**: staircases and the beam that holds each one up. Destroying the beam collapses the stair.
- **`TileApplier` / `TileFloorSupport`**: how building a floor tile in open space works. It spawns an invisible support marker that owns the tile; removing the marker removes the tile.
- **`StructuralScanner`**: a handheld tool that draws a client-side heat-map (red for roof about to collapse, green for stable build ground above a beam).
- **`PlayerBuilt`**: applied to anything a player constructs. Saved Builds uses it to determine what a player is allowed to save.

Debug and admin console commands: `au_zsupport` (recompute and report support state), `au_multiz` (see below), and `buildsave` / `savebuild` / `placebuild` for the saved-build flow.

### How the generated recipe items work

This is the system for adding entries to the construction menu in-game. It is gated behind the **Host** admin flag (`CustomConstructionMenuSystem.RequiredFlag`).

Recipes can be added two ways:
- Right-click any world entity, then **Construction > Add / Change Recipe / Remove**.
- In the construction menu, the **Admin Tools** panel has a Construction Items Editor, a Tiles Editor, and a Lathe Editor, and each recipe's detail panel has **Change Recipe** and **Remove Item** buttons.

Each action writes a normal, self-contained prototype file into `Resources/Prototypes/_AU14/CustomConstruction/Generated/`. Because it is a real file in the repo, it is committed, shipped to clients, and loaded on the next restart. Add writes the file, Remove deletes it, Change rewrites it. This is restart-applied by design, so additions are persisted in the codebase rather than lost on reboot.

Each generated file carries header comments (`# entity:`, `# spawnlist:`, `# category:`, `# steps:`, `# deconstruct:`, `# health:`) so the editor can read them back and pre-fill when editing. A recipe is an ordered list of steps: material stacks (`CMSteel`, `CMPlasteel`, and so on), custom material/tool entities, or tool qualities. You can also set the built entity's health, its deconstruction tools, and which spawnlist and category it appears under. On startup, a validation pass deletes any generated recipe whose referenced entity or material no longer exists, so a broken entry cannot wedge the menu.

**Remove Item** also handles vanilla recipes: it records the recipe id in a generated `au14MenuOverrides` prototype that the menu reads at population time and skips.

### Disabling multi-z on a map

There are three ways, depending on whether you are a mapper, a coder, or an admin in-round.

**Mapper (permanent, in the map prototype):**
```yaml
- type: entity
  id: MyMapThatShouldStaySingleZ
  # ...
  components:
  - type: ZBuildableMap
    enabled: false
```
With `enabled: false`, the system will not generate a level below or dig under that map, keeping it single-z. This is how a map like the Almayer is kept single-z.

**Coder (globally, all maps):**
```csharp
ZLevelBuildingSystem.GloballyEnabled = false;
```

**Admin (live, in-round) via console, requires the Mapping flag:**
- `au_multiz` lists every map and the global switch as `Multi Z-Level: Yes/No`.
- `au_multiz <mapId> on|off` toggles a single map.
- `au_multiz global on|off` toggles the whole feature.

The console toggle is networked, so the build-ghost check and the cave-in warning respond immediately, but it is not saved. To make it permanent, add the `ZBuildableMap { enabled: false }` component to the map prototype. The console command is for live inspection and override; the component is the persistent setting.

### Changes to existing (non-`_AU14`) files

These upstream files were modified. Each change is marked in-code:

- `Content.Client/Construction/UI/ConstructionMenu.xaml.cs`: added two members to `IConstructionMenuView` (`SelectedSpawnlist`, `SetSpawnlists`) so the classic and improved menus can share one presenter. The classic menu stubs them.
- `Content.Client/Construction/UI/ConstructionMenuPresenter.cs`: made the presenter view-agnostic and added spawnlist filtering, hidden-recipe filtering, and the grouped-grid population path.
- `Content.Client/Options/UI/Tabs/MiscTab.xaml` and `.xaml.cs`: added the Construction Mode (Improved / Default) dropdown in Options > Misc.
- `Content.Client/UserInterface/Systems/Crafting/CraftingUIController.cs`: selects which menu to build based on the CVar.
- `Content.Server/Construction/ConstructionSystem.Graph.cs`, `.Initial.cs`, `.Interactions.cs`: hook `MarkBuilt` (records who built each structure, used by Saved Builds, and plays the build sound) and handle the new non-consuming `entityId` step.
- `Content.Shared/CCVar/CCVars.Interface.cs`: three new CVars: `ui.construction_menu_improved` (default true), `construction.menu_add_default_material` (`CMSteel`), and `construction.menu_add_default_amount`.
- `Content.Shared/Construction/Steps/ConstructionGraphStepTypeSerializer.cs`: registers the new `entityId` construction step.
- `Content.Shared/Verbs/VerbCategory.cs`: new `Construction` verb category for the admin verbs.
- `Content.Shared/_RMC14/Construction/Prototypes/ConstructionPrototype.cs`: added a `spawnlist` DataField (controls which left-tree group a recipe appears under; empty defaults to AU14).
- `Resources/Locale/en-US/_CMU14/escape-menu/ui/options-menu.ftl`: strings for the Construction Mode option.
- `Resources/Locale/en-US/construction/steps/arbitrary-insert-construction-graph-step.ftl`: strings for the new step's guidebook entry.
- `Resources/Prototypes/Entities/Objects/Decoration/mining.yml`: wooden mine supports now also act as z-level vertical pillars (`StructuralSupport`, span 2), so they hold up roofs.
- `Resources/Prototypes/_RMC14/Entities/Structures/Machines/Lathe/lathe.yml` and `armylathe.yml`: added the AU14 auto/army lathe recipe packs to `staticPacks`, and the autolathe now accepts Plastic as an insertable material.
- `Resources/Prototypes/_RMC14/Recipes/Construction/structures.yml`: the five vanilla metal/plasteel/wood barricade recipes were given `hide: true` so only the AU14 ghost-placeable versions appear in the menu. The engineering-kit radial is unaffected.
- `LICENSE-AdvancedAtkinsonatorv2.md`: new proprietary license covering the author's original work in this PR.
## Requirements

**Changelog**

:cl:
- add: Jerrie's Mod construction style menu. The classic menu can be re-selected in Options > Misc. The classic menu does not have certain features supplied in the Improved construction menu.
- add: Saved Builds. Select structures you built, save them to a file, and place them again later. Only works on entities you built. You can either click the entity or append it by range. Saved builds' files can be shared.
- add: Z-Level building. Dig down into generated caverns and build floors.
- add: OSHA safety requirements for z-level building. Goodluck.
- add: Build partners. Allow another player to include your builds in their saves.
- add: Construction Item Editor, allows you to add entities into the construction menu.
- add: Tiles Editor, allows you to add tiles into the construction menu(for z-levels)
- add: Lathe Editor, allows you to add entities into Autolathe and Armylathes.
- add: Mass Entity Editor, allows you to add entities en-masse into the construction menu.
- add: Z-Level toggle, allows you to allow/ban z-level building(made by this PR) from prototypes and maps. Mapper's boolean to enable/disable overrides this.
- tweak: "Look Up" now allows you to first faintly look at upper z-levels, with the second time using it makes you fully look & aim up. No more "why is that empty space shooting me".
- tweak: Metal, plasteel, and wood barricades now place as a construction ghost instead of an instant build.
- admin: In-game editors for adding menu items, tiles, and lathe recipes, plus per-recipe Change and Remove, under Admin Tools (Host flag by defualt).
- admin: au_multiz console command to list and toggle which maps allow vertical building.
- admin: There are 3 different modes for saving & loading saved builds, Player, Admin, and Mapper. Player spawns construction ghosts by as if its vanilla, and can only save entities they built themselves. Admin spawns the saved build instantly without any cost. Mapper does the same as Admin, except they are able to save any and all entities spawned in the round.
- code: The construction menu now runs through one view-agnostic presenter with a spawnlist system, a new entityId construction step, and a Construction verb category.
- bug: The support beam construction ghost sprite appears as the 4-tile large version. This is not real. You are seeing things. Building it spawns the 1-tile version.
- add: INSFOR Editor. INSFOR factions can now be made. Every detail about what makes an INSFOR faction is covered.
- add: CLF Leader can now pick which INSFOR faction they want. It is dependent on the GOVFOR faction in the round.
- add: New Role! 🔧🧰 CLF Sapper. Able to lay down traps, craft them from the sapper's workbench, and gunsmith-lite. Equipment include: [Tripwire, Audio, Buried *#&, Snare, Spy Camera, 💵 Siphon Rig 💵 ]
- add: Siphon Rig, able to siphon from colony budget console, ASRS consoles, and ATMs. ATM siphon is based on how many ATMs are in the round(i.e 10 ATMs would give you 10% of the colony budget)
- add: CLF Sapper's Gunsmithing. Able to forcefully attach any attachments into a weapon.
- add: A switch(attachment). Craftable in the CLF Sapper's Workbench. Horrible decision to use it.
- add: Added "A Package". A box of contents(set by the INSFOR Editor) that selected jobs receive based on the INSFOR faction, if set by the INSFOR Editor(s).
- tweak: INSFOR Editor function added to Improved Construction Menu
- tweak: CLF Leader can now individually pick the items from the heavy cell kit to place down.
- tweak: INSFOR/CLF can now use captured WS equipment.
- tweak: Each 1 construction skill point reduces material consumption(metal/plasteel/wood) and increases building speed by 10%.
- remove: Removed the restriction on placing down barricade ghosts and foldable barricades for the ungifted in construction.
- add: Z-Sync List, admin menu for controlling which wall prototypes sync across z-level borders.
- add: z-sync whitelist/blacklist persistence, parent-aware matching, blacklist-family expansion, and in-round "pick clicked entity" whitelist/blacklist buttons
- add: Mass Tile Recipe mode to the Mass Entity Editor, including tile search, tile previews, shared cost/category form, and server-side batch tile recipe generation.
- add: Client now has "Faint Look Up opacity" slider in their options, clamped from 5% to 80%.
- add: Separated HUD's viewport now can be centered. Yes.
- add: Collapse vignette feedback for nearby z-level collapses/cave-ins
- add: (EXPERIMENTAL) multi-z saved-build support: saves/places entities across linked z-levels and creates missing target levels when possible.
- add: Material shortfall tracking so construction skill discounts cannot be converted into extra refund materials on deconstruction.
- add: Added eyewear, masks, and headwear to CLF and colony/WeYu role loadouts
- tweak: Separated HUD chat layout can now be shrinked further
- tweak: z-level generated maps to use map-grid entities instead of child grids, improving CMU z-level movement/tile detection
- tweak: cave-in/collapse camera shake from map-wide to local range-limited feedback
- tweak: saved-build names to mark multi-z saves with (Multi-Z). Once again, heavily experimental. Best save per z-level to be sure of your saves.
- tweak: CLF Sapper radio prefix to CIV
- tweak: RobustToolbox `Splitcontainer` with a free resize mode used by the Separated HUD chat
- admin: Added Z-Sync lists button to the construction's admin tools menu.
- admin: Added mass tile construction recipe tool.
- remove: Removed forced minimum chat width behavior in Separated HUD
- remove: Removed map-wide collapse shake behavior


Content.Server/_AU14/ZLevelBuilding/TileApplierSystem.cs
```diff
@@ -0,0 +1,93 @@
...
+    /// <summary>When a tile's support marker dies (collapses), remove the floor tile it stood for.</summary>
+    private void OnSupportTerminating(Entity<TileFloorSupportComponent> ent, ref EntityTerminatingEvent args)
+    {
+        var xform = Transform(ent);
+        if (xform.GridUid is not { } gridUid || TerminatingOrDeleted(gridUid) || !_gridQuery.TryComp(gridUid, out var grid))
+            return;
+
+        var tile = _map.TileIndicesFor(gridUid, grid, xform.Coordinates);
+        _transform.Unanchor(ent.Owner, xform);
+        _map.SetTile(gridUid, grid, tile, Tile.Empty);
```
TheHellFireo: **Do not mutate grid topology from a termination callback.** This path is currently crashing three CI jobs. Entity termination is already traversing lifecycle state; removing or replacing tile/support data from here re-enters map/entity mutation at an unsafe point. Record the affected coordinates and process them after termination, or move the invariant into the authoritative tile-change path. Add a regression test that deletes a support while dependent tiles/entities exist.

Content.Server/_AU14/Insurgency/InsurgencyFactionValidator.cs
```diff
@@ -0,0 +1,121 @@
...
+            j.Icon is not { } jobIcon ||
+            !prototypes.HasIndex<FactionIconPrototype>(jobIcon.Id));
+
+        PruneEntities(definition.CellKit.PlaceableEntities, prototypes);
+
+        // Drop submittable-for-points entries whose entity the server does not know.
+        definition.Economy.PointsSubmissions.RemoveAll(s =>
+            string.IsNullOrWhiteSpace(s.Entity.Id) || !prototypes.HasIndex<EntityPrototype>(s.Entity.Id));
+
+        // A vendor whose base model is unknown could never spawn, so drop the whole definition.
```
TheHellFireo, line 155: **Validate the complete object graph, not only the faction shell.** The nested sections that later drive vendor inventory, spawn counts, packages, and points are not recursively constrained here. A client can submit negative values, excessive quantities, or arbitrary nested content that survives this validator. Validate every collection and numeric bound server-side, then compile the draft into an immutable runtime plan.

Content.Server/_AU14/Insurgency/InsurgencyFactionValidator.cs
```diff
@@ -0,0 +1,121 @@
...
+
+        // Drop loadouts whose role is not a real job; their package could never be matched anyway.
+        definition.RoleLoadouts.RemoveAll(l =>
+            string.IsNullOrWhiteSpace(l.Role) || !prototypes.HasIndex<JobPrototype>(l.Role));
+
+        return definition;
+    }
+
+    // Removes any entity id the server does not know from a placeable/contents list in place.
+    private static void PruneEntities(List<EntProtoId> list, IPrototypeManager prototypes)
```
TheHellFireo, line 179: **Prototype existence is not authorization.** Any prototype present in the repository is accepted, including content that was never intended for player-authored factions. Use explicit server-owned catalogs or `EntityWhitelist`/`EntityBlacklist` constraints per field; reject everything outside those lists.

Content.Server/_AU14/Insurgency/Editor/InsurgencyFactionEditorEui.cs
```diff
@@ -0,0 +1,259 @@
...
+    private static string SanitizeFileName(string name)
+    {
+        foreach (var ch in System.IO.Path.GetInvalidFileNameChars())
+            name = name.Replace(ch, '_');
+        return name;
+    }
+
+    private async void HandleSave(InsurgencyFactionSaveMessage msg)
+    {
+        var def = InsurgencyFactionValidator.Sanitize(msg.Definition);
```
TheHellFireo: **Do not use a weaker validator for the custom-editor path.** This path can create the same live faction state but bypasses the stronger checks used elsewhere. All ingress routes must call one authoritative validation/compilation service; otherwise security depends on which UI submitted the same data.

Content.Server/_AU14/Insurgency/InsurgencyFactionApplySystem.cs
```diff
@@ -0,0 +1,358 @@
...
+            if (definition.Invulnerable)
+                _godmode.EnableGodmode(vendor);
+
+            var baseMarker = EnsureComp<InsurgencyFactionVendorComponent>(vendor);
+            baseMarker.VendorIndex = index;
+            return;
+        }
+
+        var comp = EnsureComp<CMAutomatedVendorComponent>(vendor);
+        comp.Sections = definition.Sections;
```
TheHellFireo: **Do not retain mutable client DTO collections as runtime state.** Assigning these lists aliases data whose ownership and validation lifetime are unclear. Deep-copy and normalize into immutable server-owned structures after validation so later mutation cannot silently change live vendor/faction behavior.

Content.Server/_AU14/Insurgency/InsurgencyFactionApplySystem.cs
```diff
@@ -0,0 +1,358 @@
...
+        _activeFaction?.Economy.DollarsToPointsRate ?? FactionDefinition.DefaultDollarsToPointsRate;
+
+    /// <summary>
+    ///     Apply a faction definition for the round. Server-authoritative: callers pass an already
+    ///     validated definition (Default factions from the DB, Custom factions after the Phase 2
+    ///     server validation gate). Safe to call again to switch factions mid-setup.
+    /// </summary>
+    public void ApplyFaction(FactionDefinition definition)
+    {
+        _activeFaction = definition;
```
TheHellFireo: **Application must be idempotent and late-join safe.** This one-shot mutation only affects entities present when the faction is applied. Vendors, players, or packages created afterward miss the configuration. Store the compiled faction plan as authoritative state and apply it from entity/player lifecycle hooks, with tests for late joins and respawns.

Content.Server/_AU14/SavedBuilds/SavedBuildSystem.cs
```diff
@@ -0,0 +1,1044 @@
...
+    /// steps from <paramref name="baseMap"/> and a grid on it under <paramref name="worldPos"/>.</summary>
+    private bool TryResolveLevel(EntityUid baseMap, EntityUid baseGrid, int zOffset, Vector2 worldPos, out EntityUid levelGrid, out MapId levelMapId)
+    {
+        levelGrid = default;
+        levelMapId = MapId.Nullspace;
+
+        var currentMap = baseMap;
+        var currentGrid = baseGrid;
+        var step = Math.Sign(zOffset);
+        for (var i = 0; i < Math.Abs(zOffset); i++)
```
TheHellFireo, line 422: **Bound Z before iterating or allocating.** Z originates in serialized/client-provided build data and this loop scales directly with it. A hostile or malformed build can force excessive work or placement far outside the intended volume. Enforce a small server-side maximum span and reject invalid bounds before preview or placement.

Content.Shared/_AU14/SavedBuilds/SavedBuildEvents.cs
```diff
@@ -0,0 +1,229 @@
...
+    public bool IncludeTiles;
+}
+
+/// <summary>One entity in a build's placement preview: prototype + position relative to the anchor.</summary>
+[Serializable, NetSerializable]
+public struct BuildPreviewEntity
+{
+    public string Proto;
+    public float X;
+    public float Y;
```
TheHellFireo, line 111: **The player placement request discards Z.** The format advertises 3D builds, but this request sends only a 2D coordinate. Multi-level previews and placements therefore cannot agree, and reconstructed entities can land on the wrong level. Either include an explicit level transform in the request or keep player blueprints strictly 2D and use the map serializer for mapper prefabs.

Content.Server/_AU14/ZLevelBuilding/ZLevelBuildingSystem.cs
```diff
@@ -0,0 +1,766 @@
...
+
+    /// <summary>Whether the building overhaul is allowed to operate on the given map.</summary>
+    public bool IsEnabledOn(EntityUid mapUid)
+    {
+        if (!GloballyEnabled)
+            return false;
+
+        // Default-on: only a component that explicitly says enabled:false opts the map out.
+        return !TryComp<ZBuildableMapComponent>(mapUid, out var settings) || settings.Enabled;
+    }
```
TheHellFireo, line 94: **Do not enable this feature on every map by default.** Adding the component during map initialization makes opt-out dependent on a later race/order-sensitive removal and contradicts the stated mapper-control model. Require an explicit map prototype/component opt-in, especially for protected maps such as the Bush.

Content.Server/_AU14/ZLevelBuilding/ZLevelToggleSystem.cs
```diff
@@ -0,0 +1,186 @@
...
+            session);
+    }
+
+    private void ApplyToMap(EntityUid mapUid, bool enabled)
+    {
+        var comp = EnsureComp<ZBuildableMapComponent>(mapUid);
+        if (comp.Enabled == enabled)
+            return;
+
+        comp.Enabled = enabled;
```
TheHellFireo, line 144: **Mapper denial needs an immutable precedence rule.** This toggle can override map-authored policy, so the claimed mapper opt-out is not actually authoritative. Separate `AllowedByMap` from temporary runtime enablement and make every command/admin path refuse to enable when the map policy denies it.

Content.Server/_AU14/ZLevelBuilding/ZLevelBuildingSystem.cs
```diff
@@ -0,0 +1,766 @@
...
+        // ramps at the same spot and flung the player between z-levels.
+        SpawnAnchoredOnce(TraversalStair, below.Comp.StoneGrid, belowGrid, landingTile);
+
+        if (_gridQuery.TryComp(sourceGridUid, out var sourceGrid))
+        {
+            var sourceTile = _map.TileIndicesFor(sourceGridUid, sourceGrid, sourceCoords);
+            _map.SetTile(sourceGridUid, sourceGrid, sourceTile, Tile.Empty);
+        }
+
+        _transform.SetMapCoordinates(digger, landingCoords);
```
TheHellFireo, line 449: **This duplicates the existing CMU Z-level abstraction.** Reimplementing level movement and coordinate conversion here creates two sources of truth for the same world topology. Extend the existing CMU Z-level service and make construction consume that API so movement, lookup, prediction, and support logic share one model.

Content.Server/_AU14/ZLevelBuilding/ZLevelSupportSystem.cs
```diff
@@ -0,0 +1,671 @@
...
+        var previous = new Dictionary<Vector2i, bool>();
+
+        var query = EntityQueryEnumerator<StructuralSupportComponent, TransformComponent>();
+        while (query.MoveNext(out var uid, out var comp, out var xform))
+        {
+            if (xform.GridUid != grid.Owner)
+                continue;
+
+            var tile = _map.TileIndicesFor(grid.Owner, grid.Comp, xform.Coordinates);
+            byTile[tile] = (uid, comp);
```
TheHellFireo: **[A tile can have more than one support.** Storing a single entity per tile means a later support overwrites the earlier one; removing either can then collapse a structure that is still supported. Track a set/count per tile or derive support from a spatial query, and test overlapping supports plus removal in both orders.

Content.Server/_AU14/SavedBuilds/SavedBuildSystem.cs
```diff
@@ -0,0 +1,1044 @@
...
+            _popup.PopupEntity(Loc.GetString("saved-build-error-nogrid"), user, user);
+            return;
+        }
+
+        if (Transform(gridUid).MapUid is not { } targetMapUid)
+            return;
+
+        var anchor = ReadAnchor(root);
+        var savedTiles = ReadSavedTiles(root);
+        var placedTiles = PlaceSavedTiles(savedTiles, targetMapUid, gridUid, targetMap, rotation);
```
TheHellFireo: **Placement is non-transactional.** Tiles are changed before all entities/components have been validated and spawned, so a later failure leaves a partially applied build. Preflight the entire plan, stage operations, then commit atomically or roll back every tile/entity mutation on failure.

Content.Server/_AU14/Construction/CustomConstruction/CustomConstructionMenuSystem.cs
```diff
@@ -0,0 +1,1475 @@
...
+        // moves the entry, so we delete the old file first; same key just overwrites in place.
+        var newKey = MakeEntryKey(proto.ID, spawnlist, category);
+        var isChange = !string.IsNullOrEmpty(msg.EntryKey);
+
+        try
+        {
+            Directory.CreateDirectory(_generatedDir);
+
+            if (isChange && !string.Equals(msg.EntryKey, newKey, StringComparison.Ordinal))
+                RetireEntryFile(FilePathForKey(msg.EntryKey));
```
TheHellFireo: **Do not delete the working recipe before its replacement exists.** A validation, persistence, or prototype-publication failure after this line permanently removes the old definition. Build and validate a new version first, publish it transactionally, then retire the old version only after success.

Content.Server/_AU14/Construction/CustomConstruction/CustomConstructionMenuSystem.Database.cs
```diff
@@ -0,0 +1,200 @@
...
+    /// <summary>
+    /// Hot-loads generated prototype YAML on the server AND every connected client, and queues it for
+    /// late joiners (same channel the admin <c>loadprototype</c> command uses). This is what makes edits
+    /// apply without a full rebuild and what keeps clients in sync with DB-restored entries.
+    /// </summary>
+    private void PublishYaml(string yaml, string what)
+    {
+        try
+        {
+            _prototype.LoadString(yaml, overwrite: true);
```
TheHellFireo, line 198: **Runtime mutation of global prototype state is the wrong persistence boundary.** Prototype graphs are startup content with cross-references and caches, not a transactional user database. Prefer generating reviewed YAML outside the game. If runtime definitions are mandatory, introduce a separate versioned recipe catalog with explicit validation, rollback, and cache invalidation rather than editing prototype-manager internals.

Content.Server/Shuttles/Systems/ShuttleSystem.FasterThanLight.cs
```diff
@@ -373,6 +387,15 @@ private void UpdateFTLStarting(Entity<FTLComponent, ShuttleComponent> entity)
     {
         var uid = entity.Owner;
         var comp = entity.Comp1;
+
+        if (HasComp<ZCollapseCompromisedComponent>(uid))
+        {
+            Log.Warning($"Aborted FTL startup for structurally compromised shuttle {ToPrettyString(uid)}.");
+            RemCompDeferred<FTLComponent>(uid);
```
TheHellFireo, line 399: **[P1] Cancellation returns with stale FTL state.** This branch invalidates the transition but exits before clearing the component/state established for it, leaving the shuttle stuck or future transitions observing a phantom jump. Route every cancellation through one cleanup method and cover this branch with a regression test.

Content.Server/_AU14/Construction/AU14MaterialShortfallSystem.cs
```diff
@@ -0,0 +1,86 @@
...
+    {
+        if (_pending.Count == 0)
+            return;
+
+        for (var i = _pending.Count - 1; i >= 0; i--)
+        {
+            var (where, stackType, missing, tick) = _pending[i];
+
+            _stacksBuffer.Clear();
+            _lookup.GetEntitiesInRange(where, RefundSearchRange, _stacksBuffer);
```
TheHellFireo: **[P1] Refunds cannot be inferred from proximity on a later tick.** Another nearby construction, moved stack, merged stack, or concurrent build can be mistaken for the material consumed by this action. Track the exact stack/entity and quantity in the construction transaction, then refund that recorded delta on failure.

Content.Client/Construction/UI/GmodConstructionMenu.xaml.cs
```diff
@@ -0,0 +1,1103 @@
...
+        };
+
+        // Admin/Player build-mode toggle. Admins default to instant/free placement, but can opt into the
+        // player flow (vanilla construction ghosts + materials) - useful when admins are playing normally.
+        // Non-admins can only ever build via ghosts, so the toggle is forced on and disabled for them.
+        var adminMgr = IoCManager.Resolve<Administration.Managers.IClientAdminManager>();
+        // Host flag only — general admin is no longer enough for the construction-menu editing tools.
+        _isAdmin = adminMgr.HasFlag(Shared.Administration.AdminFlags.Host);
+        _canEditMenu = _isAdmin
+            || IoCManager.Resolve<Players.PlayTimeTracking.JobRequirementsManager>().IsWhitelisted(EditorWhitelistJob);
```
TheHellFireo, line 438: **[P1] Client-side ACL filtering cannot define authorization.** Hiding entries here neither prevents a crafted request nor guarantees that client and server evaluate the same rules. The server must enforce one capability/allowlist service; the client may mirror its result only for presentation.

Content.Server/Construction/ConstructionSystem.Graph.cs
```diff
@@ -414,6 +414,9 @@ public bool ChangeNode(EntityUid uid, EntityUid? userUid, string id, bool perfor
             var afterChangeEv = new AfterConstructionChangeEntityEvent(construction.Graph, construction.Node, previousNode);
             RaiseLocalEvent(newUid, ref afterChangeEv);
 
+            // Stamp the builder for accountability + the saved-builds whitelist (no-op for non-players).
+            _playerBuilt.MarkBuilt(newUid, userUid);
```
TheHellFireo: **[P1] Ownership tracking is coupled to one construction path.** Marking AU14 ownership directly in the vanilla graph path misses RMC radial and any other construction entry point while forcing generic construction to depend on AU14. Raise a generic construction-completed event and let the AU14 subsystem observe it; test every supported construction path.

Content.Server/_AU14/Insurgency/Package/APackageSystem.cs
```diff
@@ -0,0 +1,144 @@
...
+
+    // A CLF member who spawns after the faction was already chosen still gets their package, as long as
+    // they spawned into a roundstart CLF job (not a reinforcement or a tattoo recruit).
+    private void OnSpawnComplete(PlayerSpawnCompleteEvent ev)
+    {
+        var faction = _apply.GetActiveFaction();
+        if (faction == null || faction.RoleLoadouts.Count == 0)
+            return;
+
+        if (ev.JobId == null || !PackageJobIds.Contains(ev.JobId))
```
TheHellFireo, line 68: **[P1] Package delivery is hard-coded to current entities and is not late-join safe.** Players created after this pass never receive faction package state. Store entitlement in authoritative faction/member state and grant it from join/spawn lifecycle handling with an idempotency guard.

Content.Server/_AU14/Insurgency/InsurgencyAuthorization.cs
```diff
@@ -0,0 +1,52 @@
...
+    // The admin flag that authorizes editing Default factions and selecting Custom factions.
+    // Host-gated to match the other Improved Construction Menu tools (Construction Item / Lathe / Tiles
+    // editors all require AdminFlags.Host), so a regular admin cannot open the INSFOR editor.
+    // Change this one constant to move the gate (for example to a future HRP whitelist manager).
+    public const AdminFlags AuthorizedFlag = AdminFlags.Host;
+
+    // The Custom-faction editor (insforcustomeditor) is open to players job-whitelisted for this job
+    // via the jobwhitelistadd command - a separate, wider group than the host flag. Change this one
+    // constant to gate it on a different whitelist job.
+    public const string CustomEditorWhitelistJob = "AU14JobCLFCellLeader";
```
TheHellFireo, line 26: **[P1] A role-name string is not an entity whitelist.** The PR describes an `InsforEditor` whitelist, but this only checks an administrative role and does not constrain which entities/prototypes may be authored. Use the existing admin capability system for editor access and separate explicit server-side content allowlists for authored data.

Content.Client/_AU14/SavedBuilds/SavedBuildListSystem.cs
```diff
@@ -0,0 +1,326 @@
...
+        }
+
+        // Keep the author prefix if present so shared files keep their origin.
+        var sep = id.IndexOf("__", StringComparison.Ordinal);
+        var prefix = sep > 0 ? id[..sep] : "local";
+        var newPath = dir / $"{prefix}__{SanitizeFileName(newName)}.build.yml";
+
+        try
+        {
+            using (var writer = _resource.UserData.OpenWriteText(newPath))
```
TheHellFireo, line 200: **[P2] Rename can silently overwrite another build.** The destination name is not checked transactionally before moving/rekeying the entry. Reject collisions (including case-normalized collisions) or require explicit overwrite confirmation, and keep the server/store authoritative for the final name.

Content.Shared/_RMC14/Construction/RMCConstructionSystem.cs
```diff
@@ -46,6 +46,30 @@ public sealed partial class RMCConstructionSystem : EntitySystem
...
+    {
+        if (!AU14DiscountedStacks.Contains(stack.StackTypeId))
+            return cost;
+
+        var level = _skills.GetSkill(user, "RMCSkillConstruction");
+        if (level <= 0)
+            return cost;
+
+        var multiplier = MathF.Max(0.1f, 1f - AU14MaterialDiscountPerSkill * level);
+        return Math.Max(1, (int) MathF.Ceiling(cost * multiplier));
```
TheHellFireo: **The documented 10% discount is not implemented.** This performs ordinary cost handling without applying or defining the claimed faction discount, so behavior diverges from the PR description. Put discount calculation in one server-authoritative cost service, define rounding explicitly, and test representative costs.

Content.Server/_AU14/Insurgency/Sapper/SapperAtmHackingSystem.cs
```diff
@@ -0,0 +1,330 @@
...
+        while (query.MoveNext(out var uid, out var card))
+        {
+            if (card.AccountBalance <= 0)
+                continue;
+
+            var skim = (int) (card.AccountBalance * fraction);
+            if (skim <= 0)
+                continue;
+
+            card.AccountBalance -= skim;
```
TheHellFireo: **This creates a parallel money path without preserving economy invariants.** Directly creating/awarding value here bypasses whatever audit, limits, and accounting the normal economy path provides. Expose a single economy transaction API and record the source, recipient, amount, cooldown, and failure atomically.

Content.Server/_AU14/Insurgency/Editor/InsurgencyFactionEditorEui.cs
```diff
@@ -0,0 +1,259 @@
...
+            await _db.AddFactionAsync(def, isDefault);
+        }
+
+        Refresh();
+    }
+
+    private async void HandleDelete(InsurgencyFactionDeleteMessage msg)
+    {
+        // The built-in vanilla CLF is code-defined and cannot be deleted.
+        if (msg.Id == InsurgencyBuiltinFactions.VanillaClfId)
```
TheHellFireo: ** Built-in factions must not be deletable through the custom-content path.** This handler does not establish immutable/system ownership before deletion. Separate built-in and user-authored records at the storage layer and reject update/delete for built-ins regardless of what the client reports.

Content.Server/_AU14/Insurgency/Editor/InsurgencyFactionEditorEui.cs
```diff
@@ -0,0 +1,259 @@
...
+        SendMessage(new InsurgencyExportSheetResultMessage(bytes, SanitizeFileName(name) + ".xlsx"));
+    }
+
+    // Imports a faction from a filled-in spreadsheet. Only cell values are read (no formulas/macros),
+    // then the result runs through the full untrusted-payload validator (caps + unknown prototype
+    // stripping) before it is stored - identical trust model to a Custom faction payload.
+    private async void HandleImportSheet(InsurgencyImportSheetMessage msg)
+    {
+        FactionDefinition? parsed;
+        using (var stream = new System.IO.MemoryStream(msg.Workbook))
```
TheHellFireo, line 163: ** Bound workbook/input size before parsing.** Accepting an unbounded upload allows memory/CPU exhaustion before semantic validation runs. Enforce byte, sheet, row, column, string-length, and collection-count limits at ingress, then parse with cancellation and a clear error.

Content.Server/_AU14/Insurgency/Sapper/SapperTrapSystem.cs
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+    public override void Update(float frameTime)
+    {
+        base.Update(frameTime);
+
+        var now = Timing.CurTime;
+
+        // Arming is cheap and should be responsive, so check it every tick.
+        var armQuery = EntityQueryEnumerator<SapperTrapComponent>();
```
TheHellFireo: **Avoid a global per-tick scan for trap behavior.** This scales with every trap for every server tick even when nothing relevant changed. Use collision/trigger events, timed updates only while armed, or a spatial query around changed entities; add a load test before introducing another always-on gameplay system.

TheHellFireo (changes requested):
Putting these files under `_AU14` does not isolate them. Generic construction, shuttle, UI, serializer, and RMC code now imports AU14-specific types, so the dependency direction is backwards


- Three CI jobs currently fail in `TileApplierSystem.OnSupportTerminating`; this is introduced by the PR and is a merge blocker. [Failing run](https://github.com/AU-14/ColonialMarinesUniverse/actions/runs/29432260699)
- Client-authored faction data reaches live vendors and runtime systems without recursive validation. Checking that a prototype ID exists is not authorization.
- Saved-build Z is effectively unbounded, while player placement drops Z entirely.
- Every map is made Z-buildable by default, contradicting the stated mapper opt-out design and leaving no explicit ship exclusion.
- Structural support tracking stores one entity per tile, so colocated supports overwrite each other and can collapse valid structures.
- Editing runtime construction content removes the old recipe before the replacement has been validated and published, making failed edits destructive.
- The FTL cancellation path returns without clearing the state it has just invalidated.
- No tests were added for any of these high-risk paths.

### Claims that don't match whatever was made

- Mapper denial does not take precedence over admin/player changes.
- RMC radial construction does not consistently mark ownership.
- Faction packages are not late-join safe.
- `InsforEditor` is treated as a role string, not an enforced entity whitelist.
- The described 10% material discount is not implemented.

### Architecture changes

| Current subsystem | Replace with |
|---|---|
| Runtime construction/prototype editor | An external typed generator that emits ordinary reviewed YAML. If runtime editing is truly required, use one versioned catalog and one transactional store rather than mutating prototype-manager internals. |
| One saved-build format for players and mappers | Separate `PlayerBuildPlan` data containing validated recipe IDs from mapper prefabs using the existing entity/map serializer. |
| Parallel AU14 Z-level and support stack | Extend the existing CMU Z-level abstraction and keep one authoritative structural-integrity implementation. |
| Mutable faction DTO used directly at runtime | Validate a `FactionDraft` recursively, compile it server-side into an immutable `FactionPlan`, and expose only server-owned catalogs/allowlists. |
| Custom sapper crafting, holder, trap and money paths | Reuse the existing lathe/construction, `AttachableHolder`, trap, and economy abstractions; add thin domain rules rather than parallel engines. |
| UI-owned authorization and giant stateful windows | Keep UI as a view/controller over server-authoritative services and the existing admin capability system. |

Before another review, split this into independently reviewable PRs, remove or redesign the live prototype editor, consolidate the duplicate vertical/support systems, make faction ingestion server-authoritative, and add targeted tests for validation, Z bounds, support overlap, replacement rollback, FTL cancellation, ownership, and late-join behavior.

I would not approve the current architecture or implementation.

Content.Server/_AU14/Construction/AU14MaterialShortfallSystem.cs
wray-git: Refunds are now tied to the exact construction action

Content.Server/Construction/ConstructionSystem.Graph.cs
wray-git: Construction ownership now covers the relevant construction paths

Content.Server/_AU14/Construction/CustomConstruction/CustomConstructionMenuSystem.cs
wray-git: Recipe replacement no longer destroys the old recipe prematurely

Content.Shared/_RMC14/Construction/RMCConstructionSystem.cs
wray-git: Material and construction-time skill discounts are now properly implemented

Content.Server/_AU14/Insurgency/Sapper/SapperAtmHackingSystem.cs
wray-git: Sapper siphoning now uses the transactional/auditing handling

Content.Server/_AU14/Insurgency/Editor/InsurgencyFactionEditorEui.cs
wray-git: Built-in factions are now protected

Content.Client/_AU14/SavedBuilds/SavedBuildListSystem.cs
wray-git: Saved-build rename collisions are now rejected

Content.Server/_AU14/SavedBuilds/SavedBuildSystem.cs
wray-git: Saved-build placement now validates before committing

Content.Server/_AU14/Insurgency/Editor/InsurgencyFactionEditorEui.cs
wray-git: Custom editor saves now uses authoritative validation

Content.Server/_AU14/Insurgency/InsurgencyFactionApplySystem.cs
wray-git: Faction runtime state is now server-owned and cloned

Content.Server/_AU14/Insurgency/InsurgencyFactionApplySystem.cs
wray-git: Reapplication and late joining are now handled even more safer :godo:

Content.Server/_AU14/Insurgency/InsurgencyFactionValidator.cs
wray-git: Nested faction collections and numeric fields are now bounded

Content.Server/_AU14/ZLevelBuilding/ZLevelSupportSystem.cs
wray-git: This wasn't very possible, but regardless I went ahead and now overlapping structural supports are tracked correctly

Content.Server/_AU14/SavedBuilds/SavedBuildSystem.cs
wray-git: Saved-build z offsets and allocations are now bounded

Content.Server/_AU14/Insurgency/Sapper/SapperTrapSystem.cs
wray-git: Should use scheduled events now, don't know how I missed this initially.

Content.Server/_AU14/Insurgency/Editor/InsurgencyFactionEditorEui.cs
wray-git: Workbook and imported-input sizes are now limited, even though this includes a trusted authoritative member in the process when handling the sheets.

Content.Server/_AU14/ZLevelBuilding/TileApplierSystem.cs
wray-git: Mmm........... Alright. I wish these testcases were clarified earlier as requested, but it's fixed now. Floor removal deferred safely, and CI regression coverage is added.

Content.Server/Shuttles/Systems/ShuttleSystem.FasterThanLight.cs
wray-git: Reviewed code already called `ReCompDeferred<FTLComponent>`, so the phantom jump claim is just wrong here. The current implementation also generalizes the cancellation through `ShuttleFTLSafetyEvent` and refreshes consoles in the faster than light .cs, I didn't add a dedicated FTL regression test, but there is no stale component after the deferred-removal tick

Content.Server/_AU14/ZLevelBuilding/ZLevelToggleSystem.cs
wray-git: This is more of a design disagreement. Regardless, with the in-game editor, it's much more modular, easier & quicker to handle, alongside allows for handling of unique events cause of the modularity & flexibility.

Also some of the z-level toggling could be depreciated from older versions, will look into it later.

Content.Server/_AU14/ZLevelBuilding/ZLevelBuildingSystem.cs
wray-git: As explained earlier in the "Mapper denial needs an imm....", toggling multi-z here can be handled quickly and instantly in-game. Check out the comment there for more details.

Content.Client/Construction/UI/GmodConstructionMenu.xaml.cs
wray-git: Server handlers now enforce per-tool ckey permissions, client filtering is presentation only.

Content.Server/_AU14/Insurgency/Package/APackageSystem.cs
wray-git: The package delivery is now granted from `PlayerSpawnCompleteEvent`, retains active faction state and uses a per-mob idempotency guard.

Content.Server/_AU14/Insurgency/InsurgencyAuthorization.cs
wray-git: This was already partially fixed, as the tool access now uses ckey permissions. Current prototypes allowlists were not added though.

Content.Shared/_AU14/SavedBuilds/SavedBuildEvents.cs
wray-git: I tried storing and projecting per-z entry offsets onto linked levels, but its been unreliable. I've left a warning and documentation for how players can save across multi-z levels in a more stable, functional, manner.

Content.Server/_AU14/Construction/CustomConstruction/CustomConstructionMenuSystem.Database.cs
wray-git: Runtime prototype is fundamental to the feature, and this is somewhat of an architectural disagreement

Content.Server/_AU14/Insurgency/InsurgencyFactionValidator.cs
wray-git: Mmmm.... please make sure you double check the change request... This is pure design disagreement, and these features are mainly used by Host and those approved directly by the Host, there would be no reason for setting up whitelists.

PR #1603, opened 2026-07-13
Port the AEV from CM13 PvE

troytroy400:
https://github.com/cmss13-devs/cmss13-pve/pull/256

image

Ports the M34E1 Armored Engineering Vehicle from CM13 PvE and does some minor clean up to the available vehicle entries. Removes the ARC version of the Humvee and command version of the UPP tank as they are identical copies.

**Changelog**

:cl:
- add: Added the M34E1 Armored Engineering Vehicle. An alternative selection to the APC, it is unarmed but incredibly good at clearing enemy fortifications. 
- remove: Removed duplicate vehicle entries from the lift console.
- fix: UPP vehicles in the lift console now use lore accurate names.


GitHubUser53123:
the command tank is incredibly useful because you have another person commanding the tank instead of focusing on one of the primary functions.

PR #1677, opened 2026-07-21
Tackle priority for non-hive members

blueDev2:
## About the PR

Prevent Sprite Blocking from affecting tackles and attacks (non-wideswing) by xenos.
All xeno attacks still require sprite-clicking, but marine sprites underneath xeno sprites are prioritized.
## Technical details

When performing light and disarm attacks, if the attacker is a xeno, replace the target with the living mob entity that is closest to the mouse click. Prioritize non-hive members as targets.
## Media

Previously:

https://github.com/user-attachments/assets/a7e40a10-519a-4a4d-8cb3-b7ee1c9c1ef8

Updated:

https://github.com/user-attachments/assets/3d4464b8-59cd-4ce5-a38a-1ebafce23177

## Requirements

**Changelog**

:cl: blueDev2
- tweak: Marines are now prioritized for xeno attacks, even if they are not the top sprite from the xeno's POV.


rokudara-sen:
Is this the greatest PR to have ever been made?

Content.Client/_RMC14/Weapons/Melee/RMCMeleeWeaponSystem.cs
```diff
@@ -57,4 +68,77 @@ private void TryPrimaryHeavyAttack()
...
+    /// </summary>
+    /// <param name="attacker"></param>
+    /// <param name="clickCoords">Mouse Click Coordinates</param>
+    /// <param name="clickedEntities">All Entities that are under the clickCoords</param>
+    /// <param name="newTarget"></param>
+    /// <returns></returns>
+    public bool TryGetAlternativeXenoAttackTarget(EntityUid attacker, MapCoordinates clickCoords, List<EntityUid> clickedEntities, [NotNullWhen(true)] out EntityUid? newTarget)
+    {
+        newTarget = null;
+        var tackleableEnts = clickedEntities;
```
maertayn: Why make a fresh new list again here every click?

You already use ToList() to make a list in the method call argument.. 

Both of it is wasteful if you look at the GameplayStateBase.cs class for GetClickableEntities, which already returns an IEnumerable, and it has already done the work internally, you're just doing it again & then once more again.

Content.Client/_RMC14/Weapons/Melee/RMCMeleeWeaponSystem.cs
```diff
@@ -57,4 +68,77 @@ private void TryPrimaryHeavyAttack()
...
+    /// <summary>
+    /// Gets the closest alive mob that was sprite clicked.
+    /// Prioritizes mobs that are not hive members.
+    /// </summary>
+    /// <param name="attacker"></param>
+    /// <param name="clickCoords">Mouse Click Coordinates</param>
+    /// <param name="clickedEntities">All Entities that are under the clickCoords</param>
+    /// <param name="newTarget"></param>
+    /// <returns></returns>
+    public bool TryGetAlternativeXenoAttackTarget(EntityUid attacker, MapCoordinates clickCoords, List<EntityUid> clickedEntities, [NotNullWhen(true)] out EntityUid? newTarget)
```
maertayn:
```suggestion
    public bool TryGetAlternativeXenoAttackTarget(EntityUid attacker, MapCoordinates clickCoords, IEnumerable<EntityUid> clickedEntities, [NotNullWhen(true)] out EntityUid? newTarget)
```

Content.Client/_RMC14/Weapons/Melee/RMCMeleeWeaponSystem.cs
```diff
@@ -57,4 +68,77 @@ private void TryPrimaryHeavyAttack()
...
+                continue;
+            }
+            nonHiveMobTargets.Add(ent);
+        }
+
+        // Prioritze non-hive entities for targeting
+        if (nonHiveMobTargets.Count > 0)
+        {
+            nonHiveMobTargets.Sort(compareDistance);
+            newTarget = nonHiveMobTargets.First();
```
maertayn: Why are you sorting this? 
Also nonHiveMobTargets[0] the same but without Linq

Content.Client/_RMC14/Weapons/Melee/RMCMeleeWeaponSystem.cs
```diff
@@ -57,4 +68,77 @@ private void TryPrimaryHeavyAttack()
...
+            return true;
+        }
+
+        if (hiveMobTargets.Count > 0)
+        {
+            hiveMobTargets.Sort(compareDistance);
```
maertayn: `foundEntities.Sort(_comparer);` is done by GetClickableEntities() and here you do it again `hiveMobTargets.Sort(compareDistance);`, you don't need draw priority sort or distance sorting, you just want the closest the rest is total noise and doesn't need computation.

Content.Client/_RMC14/Weapons/Melee/RMCMeleeWeaponSystem.cs
```diff
@@ -57,4 +68,77 @@ private void TryPrimaryHeavyAttack()
...
+    /// <returns></returns>
+    public bool TryGetAlternativeXenoAttackTarget(EntityUid attacker, MapCoordinates clickCoords, List<EntityUid> clickedEntities, [NotNullWhen(true)] out EntityUid? newTarget)
+    {
+        newTarget = null;
+        var tackleableEnts = clickedEntities;
+
+        var compareDistance = new Comparison<EntityUid>((EntityUid a, EntityUid b) =>
+        {
+            var coordA = _transform.GetMapCoordinates(a);
+            var coordB = _transform.GetMapCoordinates(b);
```
maertayn: Every a <-> b comparison is calling a GetMapCoordinates()

Content.Client/_RMC14/Weapons/Melee/RMCMeleeWeaponSystem.cs
```diff
@@ -57,4 +68,77 @@ private void TryPrimaryHeavyAttack()
...
+    /// <param name="clickCoords">Mouse Click Coordinates</param>
+    /// <param name="clickedEntities">All Entities that are under the clickCoords</param>
+    /// <param name="newTarget"></param>
+    /// <returns></returns>
+    public bool TryGetAlternativeXenoAttackTarget(EntityUid attacker, MapCoordinates clickCoords, List<EntityUid> clickedEntities, [NotNullWhen(true)] out EntityUid? newTarget)
+    {
+        newTarget = null;
+        var tackleableEnts = clickedEntities;
+
+        var compareDistance = new Comparison<EntityUid>((EntityUid a, EntityUid b) =>
```
maertayn: No need for a delegate comparison obj here

Content.Client/_RMC14/Weapons/Melee/RMCMeleeWeaponSystem.cs
```diff
@@ -57,4 +68,77 @@ private void TryPrimaryHeavyAttack()
...
+                return -1;
+            }
+            else
+            {
+                return 0;
+            }
+        });
+
+        List<EntityUid> hiveMobTargets = new();
+        List<EntityUid> nonHiveMobTargets = new();
```
maertayn: Again, two lists here, filled with noise.
You really only care about 4 variables, not the whole tile(s)

bestEnemy + bestEnemyDist (non-hive)
bestHive + bestHiveDist

Content.Client/_RMC14/Weapons/Melee/RMCMeleeWeaponSystem.cs
```diff
@@ -57,4 +68,77 @@ private void TryPrimaryHeavyAttack()
...
+
+        foreach (var ent in tackleableEnts)
```
maertayn:
```suggestion
        foreach (var ent in clickedEntities)
```

maertayn (changes requested):
The design of this has issues that have tangible impact on the game and small design anomalies like this will stack overtime.

All these lists + sorting and comparisons, that's a lot of noise and this is too expensive for every click, you only care about 2 things for this, the closest enemy & closest non-hive (and even that difference is questionable -> are hive members considered targets/enemies?)

You can just do a single pass without too much computation, something like this:

```
EntityUid? closestNonHive = null;
EntityUid? closestHive = null;
float closestNonHiveDist = float.MaxValue;
float closestHiveDist = float.MaxValue;

foreach (var ent in clickedEntities)
{
    ...
    var dist = (_transform.GetMapCoordinates(ent).Position - clickCoords.Position).Length();
    if (_hive.FromSameHive(attacker, ent))
    {    
        if (dist < closestHiveDist)
        {
            closestHiveDist = dist;
            closestHive = ent;
        }
        continue;
    }
    else if (dist < closestNonHiveDist)
    ...
}
newTarget = closestNonHive ?? closestHive;
return newTarget != null;
```

Content.Client/_RMC14/Weapons/Melee/RMCMeleeWeaponSystem.cs
blueDev2: var tackleableEnts is not a deep copy, it's a reference copy.
Both clickedEntities and tackleableEnts would refer to the same List.

I personally do this to 'rename' the list to better contextualize it within the function.

I will be removing the variable anyways, but I'm pretty sure clickedEntities would take the same space as a primative.

Content.Client/_RMC14/Weapons/Melee/RMCMeleeWeaponSystem.cs
blueDev2: _comparer sorts by draw depth. 

I want the distance from the click to the mob in the x-y plane

Content.Client/_RMC14/Weapons/Melee/RMCMeleeWeaponSystem.cs
maertayn: Right right! I misread that, it's just making a reference copy, apologies

blueDev2:
So i think there are 2 situations where this retargeting is useful:
1. Multiple xenos trying to tackle a tall, and a leadered xeno accidentally tackling another xeno
2. Moving a capture past a doorway (or other obstacle that happens to have higher sprite priority), as seen in the video, can make a tackle fail as the xeno tries tackling the airlock instead of the tall.

blueDev2:
Just more of an explanation on why I'm doing things in a way that seem odd, 
I personally find 3 or more bracket deep statements to be too confusing to read, 
so I really try to avoid getting that deep if I can.

maertayn:
Yeah that makes sense, thank you for applying the one pass approach, it's exactly what I had in mind, looks much cleaner & no more per-click allocations. Apologies again for creating confusion about the var x = x, reference copies handle that just fine - so don't change your preference/style because of my remark

blueDev2:
I appreciate it, thank you for your work.

PR #1679, opened 2026-07-21
Fix queen's missing 30 minute resin boost 

deepdarkdepths:
## About the PR
Fix queen's missing resin boost that allows her to make hivelord-level resin for the first 30 minutes of the round
Forgot to mention I also buffed the range of the resin upgrade

## Technical details
Half of what made the resin boost work was baked into `CMDistressSignalRuleSystem`. We do not use this rule system anymore as it is leftover from RMC, and as an additional problem, we have multiple gamemodes, so it is inefficient to track it through those. Instead, the following is done:

`XenoSystem` detects when a queen initializes, so it doesn't matter whether a queen is admin spawned, mapped, or a xeno evolves into a queen
New `QueenBuildingBoostSystem` listens for `QueenSpawnedEvent` and applies the boost, handles removing it, and uses its own duration(the cvars probably dont work now but who's gonna use those)

Another issue is that this doesn't work for CF since it only tracks minutes via round start and not via when the queen spawns but since CF is a slower gamemode it shouldn't matter much. Mostly intended for DS

## Requirements

**Changelog**

:cl:
- fix: Fixed queen's missing 30 minute resin boost

### DIFF
--- a/Content.Server/_RMC14/Xenonids/Construction/QueenBuildingBoostSystem.cs
+++ b/Content.Server/_RMC14/Xenonids/Construction/QueenBuildingBoostSystem.cs
@@ -1,0 +1,78 @@
+﻿using Content.Server.GameTicking;
+using Content.Shared._RMC14.QueenSpawned;
+using Content.Shared._RMC14.Xenonids.Construction;
+using Content.Shared._RMC14.Xenonids;
+using Content.Shared.GameTicking;
+
+namespace Content.Server._RMC14.Xenonids.Construction;
+
+public sealed class QueenBuildingBoostSystem : EntitySystem
+{
+    [Dependency] private readonly GameTicker _gameTicker = default!;
+
+    private static readonly TimeSpan QueenBoostDuration = TimeSpan.FromMinutes(30);
+
+    private bool _boostExpired;
+
+    public override void Initialize()
+    {
+        base.Initialize();
+
+        SubscribeLocalEvent<QueenSpawnedEvent>(OnQueenSpawned);
+        SubscribeLocalEvent<RoundRestartCleanupEvent>(OnRoundRestartCleanup);
+    }
+
+    private void OnQueenSpawned(QueenSpawnedEvent args)
+    {
+        if (_boostExpired)
+            return;
+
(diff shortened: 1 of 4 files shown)


Content.Server/_RMC14/Xenonids/Construction/QueenBuildingBoostSystem.cs
```diff
@@ -0,0 +1,78 @@
...
+    {
+        var construction = EntityManager.System<SharedXenoConstructionSystem>();
+
+        var queens = EntityQueryEnumerator<QueenBuildingBoostComponent>();
+
+        while (queens.MoveNext(out var queen, out _))
+        {
+            construction.RemoveQueenBoost(queen);
+
+            Log.Info($"Removed queen building boost from {queen}");
```
maertayn: Use sawmills

Content.Server/_RMC14/Xenonids/Construction/QueenBuildingBoostSystem.cs
```diff
@@ -0,0 +1,78 @@
...
+    private void ApplyQueenBoost(EntityUid queen)
+    {
+        var construction = EntityManager.System<SharedXenoConstructionSystem>();
+
+        construction.GiveQueenBoost(
+            queen,
+            1.5f,
+            10f);
+
+        Log.Info($"Queen building boost applied to {queen}");
```
maertayn: Logger.GetSawmill("content").Info(...)

Content.Server/_RMC14/Xenonids/Construction/QueenBuildingBoostSystem.cs
```diff
@@ -0,0 +1,78 @@
...
+            return;
+
+        if (_gameTicker.RunLevel != GameRunLevel.InRound)
+            return;
+
+        if (_gameTicker.RoundDuration() < QueenBoostDuration)
+            return;
+
+        RemoveQueenBoosts();
+        _boostExpired = true;
```
maertayn, line 58: Put this one line higher before the method call, or it'll re-run removing in the next update

PR #1701, opened 2026-07-24
Warlock changes and fixes

Piticzu:
## About the PR
Fixes several long-standing bugs on the existing Warlock abilities:

- **Psychic Blast** now correctly knocks back the targets it hits.
- **Psychic Crush** now correctly charges up its area of effect based on channel time, expanding one tile per pulse up to a 3-tile radius, with the actual damage tiles matching the visible warning tiles.
- **Psychic Shield** received the largest fix pass:
  - It no longer breaks automatically after catching 10 projectiles, it stays up until its integrity is depleted or the channel is interrupted.
  - When the shield's lifetime ends naturally, every projectile still frozen against it is auto-reflected in the shield's forward arc rather than the ability letting them pass through.
  - Rockets and other trigger-armed projectiles no longer detonate on contact with the shield. They are frozen mid-flight and held harmless for the shield's duration, matching what the ability's description already implies.
  - Grenades fired from a launcher are correctly caught and frozen. Grenades that were thrown by hand pass through, since a hand-thrown grenade is not a projectile and is not something the ability was ever meant to intercept.

## Why / Balance

- Blast's knockback was not properly implemented; the ability now applies the throw it advertises.
- Crush's damage previously ignored the pulse counter and hit a fixed area regardless of how long you channelled, so the ability's core charge-up mechanic was not implemented. It now scales as intended.
- Shield's 10-projectile hard cap made it bad against sustained fire, the cap has been removed and integrity is now the only limit.
- Frozen projectiles used to be let through when the shield timed out, so the warlock got no payoff for surviving the full duration. They now auto-reflect at end-of-lifetime, turning a full-duration shield into an offensive tool as well as a defensive one.
- Rockets used to detonate on the shield instead of being frozen. Now the rocket is held mid-air with its trigger cancelled until the shield ends, at which point it resumes its normal collision behaviour (or is reflected on end-of-lifetime).
- Launcher-fired grenades slipped past the collision event chain because their throw-fixture was not created in time. A one-tick-late oriented-rectangle sweep of the shield bounds catches them and hands them to the same freeze pipeline. Hand-thrown grenades are intentionally exempt: they are throwable items, not projectiles, and freezing them would be a scope creep beyond what this PR aims to fix.

## Technical details
- Crush's AoE hit detection now uses `EntityLookupSystem.GetLocalEntitiesIntersecting(gridUid, tiles)` (the same fixture-intersection pattern the boiler's acid gas uses), so mobs are hit exactly when their physics fixture overlaps a tile the warning covers.
- Crush's early-detonate press swaps the action's `WorldTargetAction` component out for `InstantAction` while channelling, then swaps back on end. Extended the `Access(typeof(...))` attribute on `InstantActionComponent`, `TargetActionComponent`, and `WorldTargetActionComponent` to allow `CMUXenoWarlockSystem` to add/remove them at runtime.
- Shield freezes trigger-based projectiles via a shared `PreventCollideEvent` hook that adds `CMUXenoFrozenProjectileComponent` to the projectile. A server-side handler in `CMUXenoWarlockShieldCollisionSystem` then cancels `BeforeTriggerEvent` while the component is present, so rockets and other trigger-armed ordnance do not detonate on contact.
- Launcher-fired grenades are caught by a one-tick-late oriented-rectangle sweep of the shield bounds and pushed through the same freeze pipeline, since their throw-fixture is not always created in time for the standard collision event.
- Reflect now runs on the same freeze path: when the reflect condition holds, the frozen projectile is released with a randomised angle inside the configured spread instead of being consumed.
- On natural shield expiry, `FinishPsychicShield` iterates every projectile still carrying `CMUXenoFrozenProjectileComponent` and releases each one through the reflect path, so nothing frozen against the shield is silently deleted at end-of-lifetime.
- `XenoAbilityPreviewOverlay` gains a Warlock code path that reads the range from `CMUXenoWarlockComponent` (not from the action's YAML) so the preview and the server-side snap rule share one source.
- The warlock shield RSI was trimmed to a single directional frame set; rotation is now applied programmatically by the shield spawn code.
- Integration and shared tests were updated to cover the fixed handlers.

## Requirements

**Changelog**
:cl:
- fix: Psychic Blast now correctly knocks back the marines it hits.
- fix: Psychic Crush's area of effect now correctly grows with channel time and matches the highlighted warning tiles.
- fix: Psychic Shield no longer breaks after catching 10 projectiles.
- fix: Psychic Shield now auto-reflects every projectile still frozen against it when the shield's lifetime ends.
- fix: Rockets and other trigger-armed ordnance no longer explode on contact with the Psychic Shield; they are frozen mid-flight for the shield's duration.
- fix: Grenades fired from launchers are now correctly frozen by the Psychic Shield. Hand-thrown grenades still pass through, since they are not projectiles.


Piticzu:
Failed testcase pretty sure not related to warlock, it's something related to chemistry

PR #1758, opened 2026-08-01
Remove temperature requirements for certain reagents

verentr:
## About the PR
Removed temperature requirements for several newly added reagents

## Why / Balance
As I have discovered, some maps have very weird temperatures, resulting in some temperature ranges being impossible to reach and thus making certain reagents unobtainable.

## Technical details
Deleted temperature values.

## Media
N/A

## Requirements

**Changelog**
:cl:
remove: temperature requirements from the newly added reagents in #1673.

### DIFF
--- a/Resources/Prototypes/_RMC14/Recipes/Reactions/medicine.yml
+++ b/Resources/Prototypes/_RMC14/Recipes/Reactions/medicine.yml
@@ -185,7 +185,6 @@
 
 - type: reaction
   id: CMUFentanylPrecursor
-  minTemp: 350
   reactants:
     CMUOxycodone:
       amount: 1
@@ -196,8 +195,6 @@
 
 - type: reaction
   id: CMUFentanyl
-  minTemp: 400
-  maxTemp: 420
   reactants:
     CMUFentanylPrecursor:
       amount: 1
@@ -210,7 +207,6 @@
 
 - type: reaction
   id: CMUMethPrecursor
-  minTemp: 370
   reactants:
     RMCMethylphenidate:
       amount: 2
@@ -223,7 +219,6 @@
 
 - type: reaction
   id: CMUMethBase
-  minTemp: 400
   reactants:
     CMUMethPrecursor:
       amount: 1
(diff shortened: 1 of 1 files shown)


CatoChristopherMrow:
These temps are possible to reach, and I've done so multiple times.

verentr:
Yes I am aware, however some maps have temperatures which results in some of these not being obtainable. Off the top of my head, stable and shepards have temperatures that are high which affect the temperature of the solution, so when you have a stage that requires a temperature within a range (fent) or you need to cool down a solution to a certain temp (amphet), these steps are impossible to do, or you have to some cursed esoteric stuff if youre on shepards and its 3pm on a Tuesday and its a full moon. My point here is, its currently so inconsistent between maps right now and its a real damn headache to get it to work, or it just straight up doesnt.

maertayn:
Don't you use water in a beaker to cool it down to room temp?

verentr:
> Don't you use water in a beaker to cool it down to room temp?

From my understanding, to cool down a solution you must use a reaction that produces a reagent in a lower quantity than what was put in, for example a 2:3 ratio of reactants to products, would result in a solution temperature drop. The inverse of this is how you would heat up a solution. However, on some maps like stable and shepherds, to some degree, the temperature of the maps themself affect the temperature of the solution. For example, a solution on stable garrison will always be ~40000K or something similar, and trying the above method to cool the solution does straight up not work (you can see the temp of a solution with reagent goggles). On shepards, is was able to make reagents that require heat (like the crude oil and freebase) WITHOUT adding any "heating" reaction. However, the only way I could get the reaction to go (I tried using a "cooling" reagent, in same quantities which works on the USSBush) was to put them in a ChemMaster3k and make pills, this would then cause the reaction to go, OR by consuming the mixture would also cause the reaction to go (This doesnt work on Stable for some reason). TLDR, its so inconsistent right now....

maertayn:
Hmm, okay, I'd rather we tackle the inconsistency itself though.

verentr:
I am fully in agreement, id like to see temperatures and maybe chemistry as a whole expanded on, however I do not have the technical prowess to go about that, or know where even to start.

PR #1788, opened 2026-08-04
updating req to match new standard issue

HotdogMan52:
## About the PR
wypmc req has id locked NSG23's which will bind on the user taking it from vendor (req tech, MP, etc.) so swapped it out with non-ID locked variant

added xm20a2 to hazops req vendor

## Why / Balance
during hazops round, some people got the smg by accident and i couldn't help them because the xm20a2 wasn't in req but xm20a2 is supposed to be standard issue

during wypmc round, someone requested NSG23 because same reasoning. Dispensed an ID lock NSG23 which immediately binded to me so the ID lock is rendered useless. Might as well dispense the non-ID locked variant

## Technical details
yaml 2 lines

## Media
image

image

## Requirements

**Changelog**
:cl:
- tweak: added twenty (20) units of XM20A2 Pulse Rifle to HAZOPS Requisitions Vendor
- tweak: replaced the NSG23 in WYPMC Requisitions Vendor to the non-ID locked variant due to the ID lock binding to the user who vends the rifle and not the actual user

### DIFF
--- a/Resources/Prototypes/_CMU14/Economy/Vendors/hazopsvendors.yml
+++ b/Resources/Prototypes/_CMU14/Economy/Vendors/hazopsvendors.yml
@@ -571,6 +571,8 @@
       entries:
       - id: AU14WeaponSMGXMP5A6HAZOPS
         amount: 30
+      - id: AUWeaponRifleXM20A2HAZOPS
+        amount: 20
     - name: standard issue secondaries
       entries:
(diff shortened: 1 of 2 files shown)


ver0098:
might also want to add the non UBF variant of nsg to standard issue primaries section in req

HotdogMan52:
> might also want to add the non UBF variant of nsg to standard issue primaries section in req

good idea

PR #1795, opened 2026-08-05
CT Turret upgrades!!!

Sniperzero112:
### Description:
Made it so the CT kits now allow for the usage of the turret variants since i spoke about this before with troy, now i've gotten around to added them to each of the faction vendors vendors.

### Changelog

🆑

- add: Turret upgrade kit to get variants on your turrets depending which one you took!
- tweak: Changed vendors to have flame turrets instead of sniper & mini variant turrets because you got upgrade kits now!


Sniperzero112:
The fail case seems to be talking with the orginal drone operator update that seems to just fail on any push :/

PR #1796, opened 2026-08-06
feat(clf): fix earpieces, add standing orders, comms kill switch for CLF

rokudara-sen:
## About the PR

This PR contains four related changes that address the same underlying problem: CLF cells can frequently spawn without functional communications, which builds frustration

**Clipped earpieces were broken in three separate ways**

* `TryGetWearer` only checked the `jumpsuit` slot. As a result, an earpiece clipped to a vest, militia coat, webbing, or armour granted no radio access. It now resolves the wearer from the clothing slot that actually contains the garment
* The out-of-range message was sent to the transform parent of the radio source. Earpieces grant *intrinsic* radio, so the source is the player and its parent is the map, meaning the message was effectively discarded. It is now sent directly to the actor
* `RADIO CHECK` enumerated packs and headsets, but not earpieces. A cell using earpieces with a single manpack could therefore report `NOTHING HEARD` even while the entire cell was listening

**The cell now starts with communications equipment it can reliably use.** The Cell Leader can draw the communications kit alongside the Radio Operator, including through the vendor, requisitions rack, and starting crate. The number of issued relays has increased from one to two. The issued CLF manpack also starts pre-tuned to the cell and command nets rather than with an empty configuration

**Standing orders.** The Cell Leader is issued a sheet on which they can write the cell's intent and pass the word. The order is added to every cell member's character brief, including members recruited later in the round. Orders can be issued three times per round, with a ten-minute cooldown

**An admin kill switch for CLF communications.** `au14.new_comms_system_clf`, together with a `clfcomms [on|off]` command gated behind `AdminFlags.Fun`, allows admins to return CLF and INSFOR channels to stock radio behaviour mid-round without disabling the full communications system for GOVFOR and OPFOR

The communications guidebook has also been updated to reflect the new behaviour

## Why / Balance

`radioCLF` and `radioCLFCommand` both use `anchorGated: true`. A cell without anything to anchor its network does not have reduced communications; it has no communications at all. Transmissions are sent, but nothing is received

That makes each of the earpiece bugs above a complete communications failure rather than a minor inconvenience. It also makes the round-start equipment issue a core dependency rather than a balance detail

**On the round-start kit.** The Radio Operator slot is frequently unfilled. Restricting the cell's only manpack, base station, and splice kit to a role that nobody selected can leave the entire cell shouting across the map for the duration of the round. That is not a meaningful player decision, nor is it something the opposing side can interact with

Allowing the Cell Leader to draw the same kit does not add a new capability. Both roles already carry `ANPRCRadioUserComponent`, and they are already the two roles permitted to deploy a relay

The second relay prevents one poor placement or early loss from permanently ending the cell's coverage. Pre-tuning the issued manpack follows the same reasoning: this is a militia without a formal signals pipeline, and the person who picks up the pack is at least as likely to be the Cell Leader as the Radio Operator. They should not be learning the control panel at H-hour

Captured GOVFOR and OPFOR equipment is unchanged and still starts untuned

**On standing orders.** This is intended as a coordination tool, not an alternative communications system. The limits are what preserve that distinction

Orders are limited to three issues per round with a ten-minute cooldown. Both limits are tracked **per round rather than per sheet**, so obtaining another pad does not provide another allowance. Without those constraints, the feature would effectively become a private, unjammable, map-wide radio channel with extra steps, undermining the anchor-gated communications design this repository has been moving toward

Blank orders do not consume an issue, so an accidental interaction does not waste one of the limited uses

The order is deliberately stored in the character brief rather than on a physical document. It cannot be looted from a body, faxed, or discovered by a patrol searching the correct satchel. That is intentional: it represents information conveyed during a cell briefing, not a persistent intelligence item

The limit of three issues is an initial tuning value. If leaders regularly use all three within the first ten minutes, the cooldown is providing the actual constraint and the issue count is only adding friction. If leaders rarely reach three, the count may be unnecessary

Both values are exposed as `DataField`s on the component and can be adjusted in YAML without a code change. We should observe which limit actually binds in practice before changing either

**On the kill switch.** `au14.new_comms_system` is currently all or nothing, and its only runtime control is the `cvar` console command, which is gated behind `HOST` in `engineCommandPerms.yml`

An admin observing a CLF cell that cannot communicate currently has no practical way to intervene. The only available option disables the communications overhaul for GOVFOR and OPFOR as well, and even that requires host-level permissions

This switch provides a targeted escape hatch when the communications system fails specifically on the insurgent side during a live round, which directly addresses a problem that has occurred before

The CLF-specific system remains enabled by default. Nothing changes unless an admin explicitly disables it, and any change is announced in admin chat

## Technical details

**All changes are under `_AU14` / `_CMU14`.** The only modified file outside those trees is `Content.Server/CharacterInfo/CharacterInfoSystem.cs`, which gains a single call into a partial-class extension defined under `_AU14`

### Earpiece fixes

* `AccessoryHeadsetSystem.TryGetWearer` now resolves the wearer using `_container.TryGetContainingContainer` and `_inventory.TryGetContainingSlot`, accepting `INNERCLOTHING | OUTERCLOTHING` instead of hardcoding the `jumpsuit` slot. Outer clothing already supports the `Utility` accessory category in the same way as a jumpsuit, so the previous restriction served no functional purpose
* `ANPRCRangeSystem.OnSendAttempt` selects the message recipient using `HasComp<ActorComponent>(args.RadioSource) ? args.RadioSource : Transform(...).ParentUid`
* `ANPRCRadioSystem.OnRadioCheck` now includes a pass over `AccessoryHeadsetComponent`, keyed by `RadioGrantedTo`. It skips the sender and evaluates each wearer through the same `AddStation` path used for every other station type

### Standing orders

* `CLFStandingOrdersComponent` (shared) — defines `AuthorJobs`, `MaxLength`, `MaxIssues`, and `Cooldown`, all as `DataField`s
* `CLFStandingOrdersSystem` (server) — handles the verb, issuing orders, and round resets
* `CharacterInfoSystem.StandingOrders.cs` — partial-class extension that appends standing orders to the lore primer lines

### Kill switch

* `AU14CommsToggleSystem` is now the single source of truth for whether the communications overhaul applies to a channel. `Enabled` represents the master switch, while `EnabledOn(channel)` additionally checks the CLF-specific switch when the channel's `Faction` is `clf`. Systems responsible for range, garbling, callsign masking, and ANPRC transmission now query this system instead of independently caching `AU14CCVars.NewCommsSystem`
* `AU14ClfCommsCommand` uses `[AdminCommand(AdminFlags.Fun)]`, allowing ordinary event admins to access it. When invoked without an argument, it reports the current state and warns when the master switch is disabled, since the CLF-specific toggle is inert in that configuration

**The one non-obvious detail:** `ANPRCRangeSystem` must actively *clear* the cancellation state for a disabled network rather than simply returning

Both CLF channels use `tower: true`. By the time this handler runs, the upstream `RMCTelephoneSystem` has already cancelled the transmission because no communications tower is available, and CLF cells were never expected to own one

Simply returning would therefore leave the transmission cancelled and silence the cell, which is the opposite of the kill switch's intended behaviour. Clearing the cancellation allows the channel to fall back to stock radio behaviour

Jamming continues to apply in either mode because it is an RMC feature rather than part of this communications overhaul

### Guidebook

The Field Headsets page did not mention clipped earpieces and used a flat manual style that was inconsistent with the rest of the section. It has been rewritten to align with *Getting On The Net*

Additional corrections include:

* The handset cord reaches two and a half tiles (`HandsetRange = 2.5f`), not two
* The cell receives two relays, drawable by either leadership role
* The issued manpack starts pre-tuned
* Radio checks include headsets and earpieces rather than only radio packs

## Requirements

* [x] I have read and am following the [[Pull Request and Changelog Guidelines](https://docs.spacestation14.com/en/general-development/codebase-info/pull-request-guidelines.html)](https://docs.spacestation14.com/en/general-development/codebase-info/pull-request-guidelines.html)
* [x] I have added media to this PR or it does not require an ingame showcase
* [x] I have included a **brief** detail of the major changes for this PR in the changelog below
* [x] By submitting this code and/or assets, I confirm that I either own them or have provided the correct necessary licenses to use and distribute them. I agree to be fully responsible for any legal claims or issues arising from the use of these materials

**Changelog**
:cl:

* fix: CLF earpieces now work when clipped to a vest, coat, webbing, or armour, rather than only to a jumpsuit
* fix: Earpiece users are now notified when no relay is in range instead of receiving no feedback
* fix: RADIO CHECK now includes headsets and earpieces, preventing a cell with one manpack from incorrectly reporting NOTHING HEARD
* tweak: The CLF Cell Leader can now draw the cell's communications kit alongside the Radio Operator
* tweak: The CLF now receives two clandestine relay sets at round start instead of one
* tweak: The issued CLF manpack now starts tuned to the cell and command nets
* add: The CLF Cell Leader now receives a standing orders sheet. Writing the cell's intent and passing the word adds it to every member's character brief, including members recruited later. Orders may be issued three times per round with a ten-minute cooldown
* admin: Added the clfcomms command, which disables the new communications system for CLF and INSFOR without affecting GOVFOR or OPFOR
* code: AU14CommsToggleSystem is now the single source of truth for whether the communications overhaul applies to a channel

### DIFF
--- a/Content.Server/CharacterInfo/CharacterInfoSystem.cs
+++ b/Content.Server/CharacterInfo/CharacterInfoSystem.cs
@@ -83,6 +83,7 @@ public sealed partial class CharacterInfoSystem : EntitySystem
 
         var isThreatRole = mind != null && IsThreatMind(mind);
         PopulateLorePrimerLines(lorePrimerLines, jobId, isThreatRole);
+        AddCLFStandingOrders(lorePrimerLines, entity);
 
         // Check inventory and hands for JobTitleChangerComponent
         if (TryComp(entity, out InventoryComponent? _))
--- a/Content.Server/_AU14/Callsigns/AU14CallsignSystem.cs
+++ b/Content.Server/_AU14/Callsigns/AU14CallsignSystem.cs
@@ -1,3 +1,4 @@
+using Content.Server._AU14.Radio;
 using Content.Server._RMC14.Marines.Roles.Ranks;
 using Content.Server.Chat.Systems;
 using Content.Server.Radio.EntitySystems;
@@ -28,6 +29,7 @@ public sealed partial class AU14CallsignSystem : EntitySystem
     [Dependency] private IPrototypeManager _prototype = default!;
     [Dependency] private IGameTiming _timing = default!;
     [Dependency] private IConfigurationManager _config = default!;
+    [Dependency] private AU14CommsToggleSystem _comms = default!;
 
     private static readonly Dictionary<string, string> DefaultCommandWords = new()
     {
@@ -401,7 +403,9 @@ public sealed partial class AU14CallsignSystem : EntitySystem
     // handler runs first wins safely
     private void OnCallsignSpeak(Entity<AU14CallsignComponent> ent, ref EntitySpokeEvent args)
     {
-        if (!_commsEnabled || args.Channel == null || string.IsNullOrEmpty(ent.Comp.Callsign))
+        // the CLF nets can be dropped back to stock radio on their own, and stock radio
+        // puts a speaker on air under their own name
+        if (!_comms.EnabledOn(args.Channel) || args.Channel == null || string.IsNullOrEmpty(ent.Comp.Callsign))
             return;
 
         // the callsign only holds on the callsign factions' nets. on an open named
--- a/Content.Server/_AU14/Insurgency/Orders/CLFStandingOrdersSystem.cs
+++ b/Content.Server/_AU14/Insurgency/Orders/CLFStandingOrdersSystem.cs
@@ -1,0 +1,200 @@
+using System.Text.RegularExpressions;
+using Content.Server.Popups;
+using Content.Server.Roles.Jobs;
+using Content.Shared._AU14.Insurgency.Orders;
+using Content.Shared._CMU14.Threats.Mobs.CLF;
+using Content.Shared._RMC14.Chat;
+using Content.Shared.GameTicking;
+using Content.Shared.Mind;
+using Content.Shared.Paper;
+using Content.Shared.Popups;
+using Content.Shared.Verbs;
+using Robust.Shared.Timing;
+
+namespace Content.Server._AU14.Insurgency.Orders;
+
+/// <summary>
+///     The cell leader writes their intent on a sheet once and passes the word. From then on
+///     it sits in every cell member's character brief, including anyone recruited later.
+///
+///     Deliberately not an item drop: the orders live in the brief, not on paper, so they
+///     cannot be looted off a body, faxed, or found by a patrol searching the right satchel.
+/// </summary>
+public sealed partial class CLFStandingOrdersSystem : EntitySystem
+{
+    [Dependency] private JobSystem _jobs = default!;
+    [Dependency] private SharedMindSystem _mind = default!;
+    [Dependency] private PopupSystem _popup = default!;
+    [Dependency] private SharedCMChatSystem _cmChat = default!;
+    [Dependency] private IGameTiming _timing = default!;
+
+    private static readonly Regex Whitespace = new(@"[ \t]+", RegexOptions.Compiled);
+
+    /// <summary>
+    ///     What is standing this round, or null if nothing has been passed down yet.
+    /// </summary>
+    public string? Orders { get; private set; }
+
+    /// <summary>
+    ///     Who issued it, for the byline on the brief.
+    /// </summary>
+    public string? IssuedBy { get; private set; }
+
+    // budget and cooldown are round state, not sheet state - otherwise a second pad is a
+    // second allowance and the whole thing becomes a chat channel with extra steps
+    private int _issues;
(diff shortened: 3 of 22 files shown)


peace-lock:
Minor notes:
Thanks rokudara. 
I don't think this is enough to fix the problems of CLF organization, but a huge step
As to the optional disable of newradio, admins won't use this ever if at all. They're not even aware of the problems CLF has communicating to begin with
Since you've introduced that code, I might PR it as default myself and see if they're willing to disable CLF radio. See what people think.
This doesn't fix the main earpiece bug I don't think right? If you use handheld clandestine / ANPR it disables your earpiece until you take it off and put it back on

rokudara-sen:
> This doesn't fix the main earpiece bug I don't think right? If you use handheld clandestine / ANPR it disables your earpiece until you take it off and put it back on

I wasn't aware of this bug. I can implement a fix later today

maertayn:
Do be aware of the rebase happening soon. Check the branch on this repo.

PR #1797, opened 2026-08-06
Synth surgery up

ArturUr1:
## About the PR
Raised synthetic surgery to level 4.

## Why / Balance
There are many roles in the game with the 4th level of surgery, so that this skill can be considered truly rare. For example, surgeons govfor, surgeon CLF, scientists, the chief physician have a similar skill. And I don't really understand why a fancy and expensive machine has 3 when the rest of its skills are excellent. I don't think it will hurt much, given the main function of synthetics in helping.

I am also ready to discuss the level of surgery among colonial synthetics if it seems excessive. And yes, I apologize for the mistakes, English is not my primary language, and I used a translator.

## Requirements

**Changelog**

:cl:
- tweak: Synthetic surgery has been upgraded to level 4

### DIFF
--- a/Resources/Prototypes/_CMU14/Roles/Colony/Synthetics/AU14JobCivilianColonySynthetic.yml
+++ b/Resources/Prototypes/_CMU14/Roles/Colony/Synthetics/AU14JobCivilianColonySynthetic.yml
@@ -19,7 +19,7 @@
       RMCSkillConstruction: 3
       RMCSkillOverwatch: 1
       RMCSkillMedical: 4
-      RMCSkillSurgery: 3
+      RMCSkillSurgery: 4
       RMCSkillResearch: 1
       RMCSkillMeleeWeapons: 2
       RMCSkillPilot: 2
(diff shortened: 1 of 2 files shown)


CatoChristopherMrow:
Level 3 seems pretty standard for the construction and surgery skills for roles like this. For comparison, a CT who specializes in construction, has construction 4. A doctor, who specializes in surgery, has surgery 4. Having ONE less isn't that big of a difference, but still keeps the synth as a "Jack of all trades, master of none."

I don't really get why they would need this?
I should probably knock the govfor docs surgery down to 3 instead of 4 though, not entirely sure why they got 4 here.

ArturUr1:
Oh, good. In that case, I should close the PR.  Thank you for your reply and for clarifying the position on synthetics)

ArturUr1:
Wow

maertayn:
Synths have firearms 2 but they cannot use firearms. It's fine, synths are better than a master of one.

ArturUr1:
Thanks)

PR #1829, opened 2026-08-14
AEGIS update

maertayn:
- **✨ AEGIS**
- **🔧 upstream pickaxe to mapchecker blacklist**
- **🩹 fix XenoChargeTest not giving ChargeDelay enough time to check**
- **🩹 Fix VehicleAmmoReplenishDelay testcase mismatch**

:cl:
- add: AEGIS integration

### DIFF
--- a/.github/map_checker/matchers.yml
+++ b/.github/map_checker/matchers.yml
@@ -14,7 +14,6 @@
   - "RMCAdminFax"
   - "AdminPDA"
   - "AdminHypo"
-  - "RMCPlushie.*"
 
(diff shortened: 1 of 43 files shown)


exodusky-intrynsinc:
so we finally have aegis role or?

maertayn:
> so we finally have aegis role or?

Warships will carry a classified container, which can be unlocked as an event by high-command/administrators before the round starts (lobby) or during a round.

maertayn:
Added roles too for you.

PR #1833, opened 2026-08-15
Updates Col synth loadouts

CatoChristopherMrow:
## About the PR

Changes what the col synth can have in their pockets, and lets them use more varied equipment like their govfor counterpart.

## Why / Balance

Becuase col synth dosn't typically have access to the point buy vendors like gov synth does, I used the point loadout system to make them able to actually spend points on neat equipment. Gov synths however no longer get one free ADDITIONAL equipment ONTOP of their point vendor. I've also allowed col synths to use the policing tools pouch like gov synth, since everything in there is something that the col synth would ALREADY easily have access to. Finally, expanded the range of of available webbings both can select.

## Requirements

**Changelog**

:cl:
- add: Added most experimental tools to the col synth loadout, along with a proper point amount and cost
- remove: Removed govfor synths free extra synth equipment via loadout piece (sorrrrry)
- tweak: Expanded a few webbings and pouch options for synths

### DIFF
--- a/Resources/Prototypes/_CMU14/Roles/Colony/Synthetics/AU14JobCivilianColonySynthetic.yml
+++ b/Resources/Prototypes/_CMU14/Roles/Colony/Synthetics/AU14JobCivilianColonySynthetic.yml
@@ -110,7 +110,7 @@
 
 - type: roleLoadout
   id: JobAU14JobCivilianColonySynthetic
-  points: 7
+  points: 30 # while normally they would get 45, other things in the loadout arn't properly balanced with points right now, so we'll just steal 15 from them.
   groups:
   - ColonySynthSpecializationAU14
   - SynthBackpackAU14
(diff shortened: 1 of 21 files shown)


CatoChristopherMrow:
Oh, I see, this is a upstream UI issue making the loadout UI not update til you save. Well, even though it's not my fault I'll fix it anyone. One sec

maertayn:
What?

CatoChristopherMrow:
> What?

So, the UI and the actual tracked numbers would desync. The UI wouldn't think you actually changed your points value despite doing so, then the server would have to reject it each time. Irregardless, it's fixed now.

Resources/Prototypes/_CMU14/Roles/Colony/Synthetics/AU14JobCivilianColonySynthetic.yml
```diff
@@ -110,7 +110,7 @@
 
 - type: roleLoadout
   id: JobAU14JobCivilianColonySynthetic
-  points: 7
+  points: 30 # while normally they would get 45, other things in the loadout arn't properly balanced with points right now, so we'll just steal 15 from them.
```
maertayn, line 113: Could you explain for future me what the 45 here references, as I see a 7

CatoChristopherMrow: Ah, in reference to the actual amount of points they're intended to get. https://cm-ss13.com/wiki/Synthetic

PR #1854, opened 2026-08-22
Lobby UI changes Part 1

DoguhanOzgurAkca:
## About the PR
This PR changes the lobby UI partially. 

Parts affected

Sidebar in the lobby, including the chat.
Options menu
Ahelp
Voting
Countdown

Parts not affected.

Customization.

## Why / Balance
The UI for the "CRT", which is the default, looked bad in my opinion. I also did not like how the information was presented, so these are fixes for all of it. 

The main focus was on the aforementioned "CRT" UI. I did try to keep most of my changes working for the base version, however, my testing wasn't as through.

With the text changing, there is now an option to make it more readable. I will experiment with this more later on in Part 2. I think the non-lobby text should probably use a more plain font and such. 

## Technical details

There is a small issue with the simplified font option where the old text are slightly misaligned. This was fixed sometimes, sometimes it wasn't, but it isn't a large issue, as the misalignment is only for old text, and only minor. 

AI was used throughly in this PR. I looked through a lot of it myself, but I have an agent that is pretty good at this stuff overall, even if it is their first time working with this repo specifically. 

Probably with part 2, I will add an option to minimize the timer, which is not able to be turned off right now.

![IMPORTANT]!

I asked a few people to give me texts for this, but they have not. 

image

In the new join UI, I wanted to give brief explanations about what each 'side' would be like, as I am not too familiar with every options CMU might have. Before the PR is merged, these sections should be populated. 

## Media

image
image

image

image

image

## Requirements

**Changelog**

:cl:
- tweak: rework the lobby UI partially.


maertayn:
> In the new join UI, I wanted to give brief explanations about what each 'side' would be like, as I am not too familiar with every options CMU might have. Before the PR is merged, these sections should be populated.

Or remove the Lorem Ipsum?

Content.Client/Lobby/UI/LobbyCharacterPreviewPanel.xaml.cs
```diff
@@ -1,29 +1,66 @@
...
+        // that rule on specificity.
+        if (StyleNano.CrtUiEnabled)
+            IgnoreAllegianceButton.Label.FontOverride = StyleNano.GetCrtFont(_resourceCache, 10);
+
+        // Base menu only: toggled on turns the button solid red. The default Pressed state for a
+        // plain Button is a thin highlighted border, easy to miss at a glance for something that
+        // changes spawn behaviour. The CRT theme already themes every button consistently through
+        // CrtButton, so this is left alone there.
+        if (!StyleNano.CrtUiEnabled)
+            IgnoreAllegianceButton.AddStyleClass(StyleNano.StyleClassButtonToggleRed);
```
maertayn, line 50: What is with these comments, can you tell your AI it doesn't need to put its monologues in code for one-liner single if statements like this?

It's adding so much complexity (and wasting tokens for you) when it's literally just adding a styled button lol..

Content.Client/Lobby/UI/CrtLobbyTheme.cs
```diff
@@ -56,7 +56,23 @@ private static void ApplyControl(Control control, bool useCrtTypography)
...
+                    AddClass(button, StyleNano.StyleClassCrtButton);
+
+                // Centring a button label takes two things, and the stylesheet can only do one of
+                // them. AlignMode centres the text inside the Label's own box; this makes that box
+                // actually span the button. Without it the Label is exactly text-width, AlignMode
+                // has nothing to centre within, and the label sits left in any button whose width
+                // comes from a MinWidth or a shared column - AHELP, CUSTOMIZE, CALL VOTE.
+                // HorizontalExpand is a plain property, not a style property, so it cannot be set
+                // from a rule and has to happen here.
+                button.Label.HorizontalExpand = true;
```
maertayn, line 74: Like, come on now, there's a 6 line block comment for a simple label expand

Content.Client/_CMU14/Interface/CrtScreenControl.cs
```diff
@@ -0,0 +1,181 @@
...
+
+        var intensity = _cfg.GetCVar(CCVars.CMUCrtEffectIntensity);
+        if (intensity <= 0f)
+            return;
+
+        var size = PixelSize;
+        if (size.X <= 0 || size.Y <= 0)
+            return;
+
+        if (_target == null || _targetSize != size)
```
TheHellFireo, line 120: GPU Memory leak,
 _target is only disposed when the control is resized, so if the windows closed, the renderer will keep this alive

Content.Client/UserInterface/Systems/Chat/Widgets/ChatBox.xaml.cs
```diff
@@ -94,7 +108,7 @@ public void SetForceLegacyPresentation(bool forced)
     private readonly Queue<RepeatedMessage> _primaryRepeatQueue = new();
     private readonly Queue<RepeatedMessage> _secondaryRepeatQueue = new();
     private readonly Queue<RepeatedMessage> _legacyRepeatQueue = new();
-    private readonly HashSet<string> _whitelist = ["mono", "scramble", "bolditalic", "bold", "bullet", "cmdlink", "color", "font", "head", "italic", "langicon"];
+    private readonly HashSet<string> _whitelist = BuildMarkupWhitelist();
```
TheHellFireo: The markup whitelist is built once and never rebuilt when the CRT theme toggles, so switching themes will leave chat using the previous mode. make it so its rebuilds the whitelist based on the ui theme and refreshes

you can probably do the same for allignments aswell

PR #1878, opened 2026-08-28
New Directional Bullet-Blocking System For Barricades

Berkin101:
## About the PR

Implement a custom directional bullet-blocking system for barricades that has a chance to block enemy projectiles from the front while allowing friendly diagonal shots from behind to pass through. No more bullets colliding with adjacent barricades while trying to shoot diagonally. Might need to tweak blockchance.

## Why / Balance

The old collisions are not reliable; you can shoot the barricade that is right in front of you at a certain angle, and don't even think about shooting through adjacent barricades because it won't work. The reason I know this is because a guy with an AT launcher (RPG) tried to shoot a xeno (runner) with an RPG while standing at the edge of a barricade and managed to hit the barricade that is right in front of him, breaking it and dying in the process, injuring 2 of his squadmates. I am surprised that there are no bug reports of barricade collisions as of yet.

## Technical details

- Uses dot product to calculate angle between projectile incoming direction and barricade front
- Barricade front defined as (0, -1) rotated by entity rotation (South-facing default)
- Negates projectile direction to calculate where the bullet comes from 

Base barricades: 180° blocking cone, 40% block chance
Sandbags: Inherits base settings
Both use BarricadeBlock for overhead shot mechanics alongside directional blocking

## Media

https://medal.tv/games/imported-clips/clips/npBKWJyffRNsi5_DS?invite=cr-MSxtQTksMTk5NDM3OTI4

## Requirements

**Changelog**

:cl:
- fix: Barricade collisions are now fixed!

### DIFF
--- a/Content.Shared/_RMC14/Barricade/DirectionalBulletBlockerComponent.cs
+++ b/Content.Shared/_RMC14/Barricade/DirectionalBulletBlockerComponent.cs
@@ -1,0 +1,20 @@
+using Robust.Shared.GameStates;
+
+namespace Content.Shared._RMC14.Barricade;
+
+[RegisterComponent, NetworkedComponent, AutoGenerateComponentState]
+public sealed partial class DirectionalBulletBlockerComponent : Component
+{
+    /// <summary>
+    /// Degrees cone in front of the barricade where bullets are blocked
+    /// </summary>
+    [DataField, AutoNetworkedField]
+    public float FrontBlockAngle = 135f;
+
+    /// <summary>
+    /// Chance to block a bullet when it's in the blocking cone (0.0 to 1.0)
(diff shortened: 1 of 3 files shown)


ararajdadasdasda:
this is bugged and doesnt let marines shoot over cades, it fucks over defenses.

maertayn:
doesn't this system already exist (reinforced barricades), if so, why not modify that instead?

Berkin101:
> doesn't this system already exist (reinforced barricades), if so, why not modify that instead?

The system already existed for all barricades (for the base barricade which is parent of all others), but it wasn't effective and it was quite buggy with collisions so I made my own.

TheHellFireo:
seems alright

PR #1886, opened 2026-08-30
Minor Greeno Creation Changes

silasadam:
## About the PR

Swapped how the greeno conversion works, now you inject the chemical into the unplanted or planted egg, which then creates the greeno facehugger responsible for making the greeno queen. Ontop of this, if a queen died recently, it will link the queens blood to the newly created greeno, giving the killed queen the option to become the greeno facehugger, or to leave it. Also reduced the hour requirement for greeno queen to the normal threat lead requirement of 5 hours, and swapped where it checks this so that you cant enter the facehugger at all unless you meet it. Added a note in the egg crate and in the ciphering sheet that says to inject the egg before letting it open.
## Why / Balance

People could troll scientists by taking the greeno facehugger without having the 30 hours required for greeno queen, grinding everything to a halt. Ontop of this, 30 hours was absolutely insane for an hour requirement, especially for a role that is generally a bit more stable than DS xeno queen or other queen roles. Furthermore, the creation of a xeno-queen just to create the greeno-queen often condemned a player to simply joining a role in order to die, and now instead of that, you're actually selecting an applicable role. [greeno queen], because you're given the option to either take the greeno-queen resultant from your death, or leave it for others to take.
## Technical details

Swapped around the test/check comps and created a quick UI popup tied to the most recently killed queen [this COULD create bugs, but its highly unlikely.]
## Requirements

**Changelog**
:cl: PUPPET/MERIDIA
- tweak: Changed greeno creation pathway
- tweak: Reduced greeno queen hour req to 5 hours of threat member [same as regular xeno queen]
- tweak: Allowed the queen murdered to make greeno queen the option to become greeno queen first.


deepdarkdepths:
Requirement reduction should be at least 10-15 not 5. This is still a HRP role whether it's more stable than DS queen or not.

Kadeo64:
-There is no reason to change how the initial chemical injection works. Injecting into a dormant egg is whatever (i'd probably keep this) but there is literally zero situation where you should be injecting into a planted egg that will automatically infect you unless you are griefing. It's kind of just bloat?
-Linking queen's blood to the greeno queen is [a mald change] unnecessary? What happens if a yung XX xeno with no threat hours (not enough for greeno queen) becomes first queen? I don't necessarily have anything against this change but i do want to know if this very possible situation was tested for.
-Not being able to enter the facehugger means that the egg could potentially collapse if scientists wait for a greeno para to pop out. While unlikely, this means that if no greeno-eligible players are online scientists could potentially lose their only shot from the egg collapsing, whereas with the previous method you could just ask the greeno to ghost in looc once you can get a high enough hour player online. (proximity auto-infect shouldn't be bugged so that method could still work.. although then you get the issue of the player potentially not having enough hours to evo to greeno queen anyway)
-the 30 hour timelock was chosen for a reason, we've been over this. I could understand lowering it a little bit (20 hours?) but five hours is still too short, and again, insurgency and DS are vastly different and you're not going to gain any meaningful queen experience in insurgency. Roleplay experience _maybe_ but you could get that as a non-queen greeno.

maertayn:
Not every insurgency round should cater to greenos

maertayn:
There's discussion to remove it altogether from Ins, so lowering the entry barrier is completely changing course with Greeno feedback and the overall plan for it

PR #1903, opened 2026-09-07
fixes dropship flight movement

CatoChristopherMrow:
## About the PR

Makes it so that the hud properly activates when you use it, instead of only on equiping the helmet itself. Also makes it so that the hud itself isn't required for moving the ship.

## Requirements

**Changelog**

:cl:
- fix: Fixed the pilot hud and the dropship movement


maertayn:
Weird testcase, not sure if you need that
Have a little more faith in the engine, it won't randomly break down and forget how components work

CatoChristopherMrow:
> Weird testcase, not sure if you need that Have a little more faith in the engine, it won't randomly break down and forget how components work

I'm used to coding in ss13. I trust nothing about engines.

PR #1908, opened 2026-09-07
Synth DR 

ararajdadasdasda:
## Why / Balance

Synths had a DR in cm13 and for some reason working joes have it, but not the synths? figured i'd leave it roughly the same, with synths having slightly better shock/heat resistance. They also can no longer use guns, even if their hacked or a runaway. 
## Technical details

All it does is add a damage modifier for synths to make them tankier like their working joe counterparts. For some reason, when synth's were made they did not have this damage resistance or it was not ported. 

## Requirements

**Changelog**

:cl:
- add: synth's have DR like they do in cm13 now, slightly better then working joes.

### DIFF
--- a/Resources/Prototypes/_RMC14/Damage/synth_modifier_sets.yml
+++ b/Resources/Prototypes/_RMC14/Damage/synth_modifier_sets.yml
@@ -14,6 +14,6 @@
     Blunt: 0.5
     Slash: 0.5
     Piercing: 0.5
-    Heat: 0.8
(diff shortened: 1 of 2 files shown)


maertayn:
Changes outside of the Content.CMU zone require // CMU14 tagging.

That way when we pull a downstream merge, your changes aren't reverted against the new diff.

More details IN CONTRIBUTING & CONVENTIONS root markdown files.

PR #1931, opened 2026-09-11
Major Yautja Update Part Three!

silasadam:
## About the PR

Returned the yautja's cloak to its 'true cloak' state [when standing still, they are completely invisible], in addition, made sure they cannot be right-clicked on when they are invisible. In addition to the earlier added 50/50 tackle roll, made it take 4 tackles or more to actually knockdown a yautja, and made tackles not cancel the 'get-up' do-after. The healing gun now has a forced do-after if a yautja has a major internal injury [Frac, IB, Eschar, etc.] But will repair it too. Yautja also now regrow limbs lost after 2 minutes and in addition, delimbing a yautja takes at least 2 successful delimb rolls. Made the yautja hunting traps completely invisible to everyone except yautja. Gave yautja a map symbol/map tracking and made yautja hunting traps temporarily show the trapped persons map symbol for about 5 minutes.
## Why / Balance

Yautja could still sorta kinda get tackle spammed. I've also been informed that for the moment we are aiming for some form of parity to CM13 predators. So for now, I've worked to make them have some kind of parity. Before this, yautja also had no way to deal with a limb being severed or things like Fracs or IB without either A: having another yautja, or B: Asking the - eugh, humans. for help. And neither of these outcomes were larpful. So I changed it. Yautja hunting traps actually have an applicable use now if the yautja leaves them as well, traps. And now they can actually kind of hunt targets along with knowing where they are without having to guess based on landmarks.
## Test plan

Load up local host -> Spawn yautja hunter -> spawn CLF soldier, spawn praetorian, spawn queen. -> CLF soldier shoots the yautja several times [For frac]. Test healing gun on yautja. -> Have praetorian tackle yautja. Should take several attempts to knock down. -> have queen hit the yautja several times and at least once with the tail stab. They should delimb maybe once every 4-6 hits from queen, and should regenerate the lost limb after about 2 minutes. -> Test 'open hunting map' ability. -> throw down hunting trap. -> Walk someone into it. -> Check map again.

## Requirements

## Breaking changes

I do not believe there were breaking changes in this PR.

**Changelog**
:cl:
- add: Added Yautja Limb regeneration
- tweak: Yautja healing gun can heal IB/Fracs/Eschar/etc on a long do-after.
- tweak: Yautja take several tackles to knock down. [4]
- tweak: Yautja hunting traps are now completely invisible to everyone except the yautja.
- tweak: Yautja cloak has been returned to true-cloak status.
- add: Yautja have a map symbol&hunting traps track the target for 5 minutes.


Content.CMU/Resources/Prototypes/CMU14/Threats/Yautja/Mobs/mobs.yml
```diff
@@ -15,6 +15,27 @@
     currentLanguage: Yautja
     defaultLanguage: Yautja
   - type: Yautja
+    xenoTackleSuccessChance: 0.5
+    xenoTackleSuccessesRequired: 4
+    xenoTackleExpireAfter: 4
```
maertayn, line 20: ~8 tackles in 4 seconds requires the whole horde to be spam tackling, combined with their 3.5% visibility. Pretty rough, can always balance it out though.

Content.CMU/Server/Yautja/YautjaTrapSystem.cs
```diff
@@ -131,6 +141,7 @@ public bool TryArmTrap(Entity<YautjaTrapComponent> trap, EntityUid user)
         trap.Comp.TrapOwner = user;
         trap.Comp.Armed = true;
         Dirty(trap);
+        SetTrapVisibility(trap.Owner, true);
```
maertayn: Odd feedback, does this match the new visibility flags?

Content.CMU/Shared/Yautja/YautjaCloakSystem.cs
```diff
@@ -30,6 +32,8 @@ namespace Content.Shared.CMU14.Yautja;
 
 public sealed partial class YautjaCloakSystem : EntitySystem
 {
+    private const float MaximumCloakedMovingOpacity = 0.035f;
```
maertayn: Why use this const float and is it not based on the bracer's (C#/yaml) movingOpacity value?

Content.CMU/Shared/Yautja/YautjaCloakSystem.cs
```diff
@@ -205,7 +213,9 @@ private bool TrySetInvisibility(Entity<YautjaBracerComponent> bracer, EntityUid
             var activeInvisibility = EnsureComp<EntityActiveInvisibleComponent>(user);
             var cloakUser = EnsureComp<ThermalCloakUserComponent>(user);
             cloakUser.Opacity = bracer.Comp.CloakOpacity;
-            cloakUser.MovingOpacity = MathF.Max(bracer.Comp.CloakMovingOpacity, bracer.Comp.CloakOpacity);
+            cloakUser.MovingOpacity = MathF.Min(
+                MathF.Max(bracer.Comp.CloakMovingOpacity, bracer.Comp.CloakOpacity),
+                MaximumCloakedMovingOpacity);
```
maertayn: Likewise, why pin it in a range and make the yaml obsolete. This is hardcoding a balance value which was originally 10% but you are now locking to a maximum of 3.5%

Content.CMU/Shared/Yautja/YautjaComponents.cs
```diff
@@ -535,10 +541,10 @@ public sealed partial class YautjaBracerComponent : Component, IClothingSlots
     public SoundSpecifier CloakOffSound = new SoundPathSpecifier("/Audio/CMU14/Yautja/Equipment/pred_cloakoff.wav");
 
     [DataField]
-    public float CloakOpacity = 0.02f;
+    public float CloakOpacity;
 
     [DataField]
-    public float CloakMovingOpacity = 0.10f;
+    public float CloakMovingOpacity;
```
maertayn: Don't remove the defaults, there's a null-value waiting to surface here. Yaml can adjust.

maertayn (changes requested):
Did you coordinate with RuCM's developer Maiji on these changes? Seems like you're not just tweaking existing values but making new systems too.

PR #1932, opened 2026-09-11
Removes the Reaper caste as a Carrier evolution

deepdarkdepths:
## About the PR
Removed Reaper as an evolution option from Carrier.

## Why / Balance
Reaper is pretty useless, and is borderline a throw pick as it takes up a T3 slot and does not contribute meaningfully to the hive. Its flesh resin mechanic relies on xenos already winning to get anything out of it. Its ability to build is overshadowed by its low plasma, so it can only get about 4-5 structures out before needing to recharge. The only good ability it has its mantle, which gives one xeno a shield, but that doesn't justify picking it.
Every other xeno does what reaper does, but better. It is meant to be a support caste, but does not do it very well. It is in dire need of a rework so I am disabling it until then.

## Requirements

**Changelog**

:cl:
no cl

### DIFF
--- a/Resources/Prototypes/_RMC14/Entities/Mobs/Xeno/carrier.yml
+++ b/Resources/Prototypes/_RMC14/Entities/Mobs/Xeno/carrier.yml
@@ -174,9 +174,9 @@
     - ActionXenoDevolve
   - type: XenoParasiteThrower
   - type: XenoEvolution
-    max: 500
-    evolvesTo:
-    - CMXenoReaper
+ #   max: 500 # CMU14 - Needs a rework before re-enabling
+ #  evolvesTo:
+ #  - CMXenoReaper
     strains:
     - RMCXenoCarrierEggsac
   - type: RMCActionOrder


maertayn:
Lower it to tier 2, don't outright remove it, even if it's not as you'd expect yet, it remains a (new) player's choice afterall


It's not a tier 3 carry I understand

deepdarkdepths:
> Lower it to tier 2, don't outright remove it, even if it's not as you'd expect yet, it remains a (new) player's choice afterall
> 
> It's not a tier 3 carry I understand

I was considering tier 2 as well but even the other tier 2s function better than it. If this pr is straight up gonna not be merged otherwise i can make it t2 though

maertayn:
Perhaps nerf the health pool so its akin to a drone support role, T1/T2, buff plasma regen, 1 specific pheromone, extra mantle stack(s)

Not sure where to take it, I generally pause and hesitate when someone's suggestion comes down to removing content

deepdarkdepths:
> Perhaps nerf the health pool so its akin to a drone support role, T1/T2, buff plasma regen, 1 specific pheromone, extra mantle stack(s)
> 
> Not sure where to take it, I generally pause and hesitate when someone's suggestion comes down to removing content

Yeah i don't like removing content either. In this case it's very little push back about removing it though.
I'll do your suggestions and open another pr later

PR #1935, opened 2026-09-12
Fix CI regressions after history reconciliation

TheHellFireo:
Repairs prototype, networking, lifecycle, UI, and regression-test failures from [run 34661509956](https://github.com/AU-14/ColonialMarinesUniverse/actions/runs/34661509956).

- Restore missing prototypes, sound sequences, localization, and moved resource references. Align serialized defaults with initialized state and retain validation of server-only prototype references.
- Serialize optional entity references safely after deletion, preserving gameplay fields and connected-client replication coverage.
- Restore static action icons after dynamic overrides clear, remove temporary icon layers on shutdown, and recreate deleted camera network seeds. World-deletion tests preserve these persistent services so cleanup terminates.
- Correct delayed solution transfers, use the consuming player's ingestion preference, and guard cleanup during state application and entity deletion.
- Restore collapsible keybind categories, including emote selectors. Searches expand matching sections and restore their previous expansion state when cleared. Respect paused autodoc and body-scanner consoles.
- Update fixtures for damage immunity, role requests, required volunteers, gravity, voices, projectile prediction, medical caches, and spawn synchronization. Hands tests use a dedicated human and tracked entity cleanup.
- Discard servers with dynamically generated reagent prototypes, preventing the flavor test's Water reaction from contaminating stain and Bosenmori map tests. Isolate lighting diagnostics that read process-wide counters. Disconnect after the synthetic follower reset before stale server updates can reach teardown, and restore pooled log thresholds so later warnings remain checked.
- Clean up the botany migration fixture's mango packet, count only seed packets created during extraction, and recycle the extractor fixture's nullspace entities. Keep species, potency, yield, and duplicate-conversion assertions intact.
- Measure failing-heart pressure through the existing origin-filtered damage probe and settle its physiology interval explicitly, preventing ordinary respiratory recovery and pooled scan timing from hiding the applied damage.
- Include the previously omitted `Content.IntegrationTests.CMU14` namespace in the CI matrix. Update stale construction, cargo, species, and lighting assertions while retaining their behavior checks.
- Raise NUnit AssemblySelectLimit to 10,000 in both CI workflows. Its default 2,000-case limit discarded the gameplay execution filter, ran the whole assembly, and hid results for cases outside the discovered group. An engine trace confirmed the mismatch; the corrected run reports gameplay results immediately.
- Split debug CI into three complementary groups with unique result artifacts and a shared success check. VSTest discovery confirms all 5,678 tests appear exactly once. Retain 90-minute job budgets, allow 75 minutes for full local runs, and enable per-test progress output; the former 30-minute job and 45-minute watchdog interrupted healthy progress.

Validation: DebugOpt content/integration builds pass, including the latest fixture changes with zero errors. Both modified workflows pass actionlint. Earlier focused checks passed for reagent/stain/Bosenmori isolation, lighting diagnostics, hands, world cleanup, keybind filtering, medical UI, ghost roles, third-party deployment, serialization, action icons, camera recreation, entity counts, Distress Signal startup, and map save/load. The complete local CMU group passed all 1,351 tests before the final fixture changes. No further local tests were run after the requested stop; subsequent execution was confined to GitHub CI.

CI on f51d276d1bf is not fully green. [Debug content](https://github.com/AU-14/ColonialMarinesUniverse/actions/runs/34720214878/job/103625219414) passed 1,667 integration tests, including the corrected botany and heart-pressure cases; its unit tests also passed (848 passed, one skipped). Debug maps/prototypes passed. Debug gameplay is still running. Every main CI group except CMU passed.

The [CMU job](https://github.com/AU-14/ColonialMarinesUniverse/actions/runs/34720214871/job/103627845810) was interrupted by runner shutdown on both attempts, with no failed assertions in either test log. The second attempt enabled runner diagnostics: job-lease renewals stopped for about six minutes before shutdown. The logs do not establish an out-of-memory error or the shutdown's root cause. This interruption also fails the Required Checks summary.

Two RobustToolbox issues remain outside this PR: parallel variant registration can corrupt its dictionary during startup, and CleanReturnAsync can publish a pair before the former borrower finishes DisposeAsync, allowing it to dispose the next test's lease. The latter also reproduced between UiLessFixOVeinRepairsInternalBleedingAtShallowOrDeepAccess and UiLessPreferenceRoutesToolClickWithoutOpeningWindow in the completed CMU run on bf87a5ac63d. RobustToolbox and its dependency reference are unchanged as requested; these intermittent races are not fixed by this PR.

This PR remains open with automatic merging disabled. It will not be merged.


flex5hybrid:
One map (hms_pratchett.yml) is failing test because it has WehMedipen that has "DO NOT MAP" tag.

PR #1943, opened 2026-09-13
Ravager strains tweaks

deepdarkdepths:
## About the PR

An assortment of value changes to the base Ravager and Berserker Ravager.
Eviscerate stun time has been increased.
Full-rage eviscerate does more damage.
Armor per rage stack has been increased.
Rage decay time reduced.
Heal per hit increased.
Fling heals for more.
Base ravager's empowered scissor cut slow duration increased, and flat damage increased.

## Why / Balance

With similar reasons to my last PR essentially, both strains were struggling with current govfor's damage output.
Base ravager just needed a little bump in damage and slow time, as its scissor cut was essentially doing nothing with our current armor values.
Berserker is a specific case, however. Berserker's entire gameplay loop functions by staying in line of fire much longer than other xenos- it needs to build stacks. It is not a peeking or get in and get out xeno, nor does it have a tool to quickly get out of an undesirable fight. This means it needs to, half the time, be out in the open and _commit_ to gaining stacks by attacking govfor. Even with fire from only a few marines, they could out-dps your full-rage healing. Without stacks however, it cannot do much of anything, as it lacks crowd control control and range tools.
Keep in mind that its hp and armor were not changed. It still remains as squishy as it was before without rage stacks. Either way, a caste like this with its unique function cannot be expected to survive when nearly all govfor weaponry and armor have much higher firerates and dps compared to parity from both rmc and cm13. 

Its Eviscerate, which is an ability that roots you, has been changed to be more rewarding to use instead of deciding whether you should risk your life for it or not.
Armor per stack was changed to make it easier to gain stacks, likewise for rage decay time.
Its healing capability was changed so it can do what it's meant to do: stay in the fight much longer than others.

## Requirements

## Breaking changes

**Changelog**

:cl:
- tweak: Base and Berserker Ravager stat values tweaked.


ararajdadasdasda:
holy mald?

PR #1955, opened 2026-09-14
Allow dropships to fly over unmapped space

Vencallium:
## About the PR

Allows dropships in tactical flight to fly over empty, unmapped space without incorrectly reporting an obstruction.

## Why / Balance

Pilots should be able to fly across clear space without needing a supporting floor tile. Walls and physical obstacles still block the dropship.

## Technical details

Updates `DropshipTacticalLandSystem.GunshipPilot.cs` so missing map chunks are treated as empty space during flight-clearance checks. Existing collision checks remain unchanged.

## Media

Small code fix; no media required.

## Requirements

## Breaking changes

None.

**Changelog**

:cl:
- fix: Dropships can now fly over empty space without incorrectly detecting an obstruction.


maertayn:
Outside the map boundaries?

Vencallium:
Ideally, all maps will have some form of rock wall to prevent people from abusing flying outside the map, but why would people go outside the map in the first place if their goal is to kill

Vencallium:
Also, unmapped space is what is used in z-level layers. It's also the only tile that allows things to drop through as well. So, for example, on Stable Garrison, open air uses the tile "space," and it's what makes things drop, but you can't maneuver through that tile or land on it.

TheHellFireo:
It causes more issues then it fixes

Vencallium:
The Lexington vtol literally does not work in its current state

PR #1960, opened 2026-09-14
Synth loadout tweaks and fix synth skill

ArturUr1:
## About the PR

The surgery skill for third-generation synthetics has been fixed. It is now 4 again. Synthetic item loadouts have also been changed. Now every synthetic can take two items regardless of what they are wearing.
## Why / Balance

Regarding the fix. While playing on the RuMC project, I noticed that third-generation synthetics now have a surgery skill of 3, even though I had changed it to 4. As it turned out, for third-generation synthetics these skills were defined in a different place. So now everything is fine.

Regarding the change to the loadout system. As it turned out, synthetic special items use customization points for selection. Because of this, a situation arises where a synthetic has to be ugly in order to have the maximum number of points for two items. I removed the point costs for the items, but limited their quantity to two. This shouldn't significantly affect balance. At least synthetics already had the ability to take two items before this. And let's not forget that all items cost 15 points, except for the purifier. Right now these prices don't make sense.
## Technical details

- The surgery skill for third-generation synthetics has been raised to 4 in a separate file.
- All prices for special items in the synthetic loadout have been removed. A maximum number of items in the loadout group has been set.
## Requirements

## Breaking changes

**Changelog**

:cl:
- tweak: The loadout system for synthetic special items has been changed. They no longer require customization points, and you can all be beautiful. However, the maximum number of special items has been set to 2.
- fix: The surgery skill for third-generation synthetics is 4 again.


UnreadLibrarian:
I appreciate this change and the spirit of it, however I would beg of you to remove the purifier / Experimental spray bottle from this category of the loadout (and thus not have it share the same budget as the rest of the special tools).
The experimental bottle will literally never ever be seen or used again (aside from Joes) if it shares a slot with the rest of the special tools.
Note: Currently the Experimental Spray Bottle costs 3 pts (as opposed to 15 for everything else in the category). This reflects the spirit of what I'm asking.

ArturUr1:
I could make a separate loadout for the purifier that wouldn't depend on the item selection, however I figured that three special items for a synth would be "unbalanced." If they approve something like that in your case, I'd be only glad to have the purifier always ;)

PR #1962, opened 2026-09-14
Tweak wall building G-menu

ArturUr1:
## About the PR

Instant wall construction through the G-menu has been removed. Now you can quickly place frames through it, and build walls through those.

## Why / Balance

I can't believe that no one noticed after all this time that these walls are built instantly and don't require any construction skills. On top of that, they also work incorrectly. First, they can be dismantled with a single crowbar hit. Second, the xenomorph side doesn't have crowbars, and also, because the wall doesn't have to be finished after being created, it gets stuck in that state, which means regular xenomorphs can't even melt it with acid. This allows either side of a conflict to build huge structures without any specialists in their respective fields.

## Technical details

All unnecessary crafting entries have been removed. The crafting of a regular metal wall has been replaced with the crafting of a frame.

## Requirements

## Breaking changes

**Changelog**

:cl:
- tweak: Instant wall construction through the G-menu has been removed. Now you can quickly place frames through it, and build walls through those.


CatoChristopherMrow:
How will I instantly entoumb unsuspecting civilians now????

PR #1973, opened 2026-09-17
Add position in queue being kept when becoming para/lesser

deepdarkdepths:
## About the PR
Larva queue position is now kept if you become a hugger or a lesser. 

## Why / Balance

QOL. Mechanic already exists in CM13.

## Test plan

Spawned paras and lessers, infected hosts, one being through proximity hug and the other through player hug. Waited for burst while playing as listed roles. 

## Requirements

**Changelog**

:cl:
- tweak: You no longer lose your position in queue when becoming a lesser drone/facehugger.


Content.Server/_RMC14/Xenonids/JoinXeno/LarvaQueueSystem.cs
```diff
@@ -690,7 +705,13 @@ private bool TryGetQueuedSession(NetUserId userId, out ICommonSession session)
         if (!_player.TryGetSessionById(userId, out session!))
             return false;
 
-        if (session.AttachedEntity is { } attached && _ghostQuery.HasComp(attached))
+        if (session.AttachedEntity is not { } attached) // CMU14
+        {
+            RemoveFromAllQueues(userId);
+            return false;
+        }
```
maertayn: This is not needed because line 717 also does this to close the boolean method.

maertayn (changes requested):

Content.Server/_RMC14/Xenonids/JoinXeno/LarvaQueueSystem.cs
deepdarkdepths: Mb, fixed

PR #1979, opened 2026-09-20
Despoiler CE 'bugfix' revert

ILoveDespoiler:
## About the PR
reverted changes made to XenoDespoilerCausticEmbraceSystem.cs which preverted it from using said ability to leap over barricades. This was reported as a bug but I believe it is intentional as the ability can be used to leap over barricades in CM13

## Why / Balance
Restores the Despoilers ability to leap over cades. You might think a Xeno being able to leap over cades is insanely strong, but really only lets Despoiler chase down singular people behind cades, which i think gives it a powerful niche.

## Technical details
Removed the lines that the fix added, re-added the lines it removed. the code should be back to its original state as before the PR

## Test plan
went into localhost. spawned barricades, spawned a despoiler, and was able to use CE to get over them
also, got onto a round of CM13 to confirm that the ability has the behavior of leaping over cades, which it does for both versions.

## Requirements

**Changelog**

:cl:
- Fix: fixed Despoiler CE being blocked by Barricades.


deepdarkdepths:
This was indeed a bug fix and was applied to Caustic Embrace because it is also a type of dash. Also, no more balance PRs.

ILoveDespoiler:
how is it a bugfix if the intended behavior of the ability, when it was originally created, allowed it to do so?

ararajdadasdasda:
I would not trust someone called "ILOVEDESPOILER" to make despoiler balance changes

PR #1984, opened 2026-09-21
Lobby localization and remove hardcode

ArturUr1:
## About the PR

The lobby localization has been transferred from the RuMC project. In particular, the English-language part. Hardcode has been replaced and localization keys have been added. The Russian translation is, of course, ours and was not transferred ;)

## Why / Balance

It was very hard for us to do upstreams and translations because of the hardcode. Now they've asked us to send this to you. We would be very grateful!

## Technical details

* Hardcode has been removed in places related to the lobby
* Localization keys have been added in the places where hardcode was replaced or in `spawnMenuRoleName`
* Localization key translations have been added in FTL format in English, comrades
## Requirements

## Breaking changes

**Changelog**

:cl:
- code: Hardcode for the lobby has been replaced. Localization keys for the roles have been added. English translations of the localization keys have been added


maertayn:
Thanks for the PR. This touches files outside the CMU-owned zones (`Content.CMU/`, `_CMU14`), and those changes aren't tagged, so they'd be lost or silently conflict on the next upstream merge.

Deliberate changes in upstream areas need a marker per contiguous block:
   - **C#** example and inline is always preferred:
     ```csharp
     public float Capacity = 10; // CMU14: colony storage capacity
     ```
     
`// CMU14 method` / `// CMU14 class` / `// CMU14 event` for new constructs.
   Larger contiguous edits use `// CMU14 <Name> Begin: reason ...` `// CMU14 End`.
 - _YAML_: `# CMU14`; a whole prototype can mark its `type:` or `id:` line.
 - _Removals_: donot just delete upstream code. A bare deletion leaves **no
   trace**, so the next upstream merge restores it without conflict. Comment out
   the removed code with a `// CMU14:` tag instead. Plain deletion is only fine
   when the enclosing construct is already fully tagged or rewritten.
 - _Exempt_: pure mechanical upstream conformance (API renames, signature
   changes needed to compile) needs no tag.

 Read [CONTRIBUTING.md](https://github.com/AU-14/ColonialMarinesUniverse?tab=contributing-ov-file#the-cmu14-tag) and [CONVENTIONS.md](https://github.com/AU-14/ColonialMarinesUniverse/blob/master/CONVENTIONS.md#fork-extensions-and-markers) for the full tagging syntax, and [AGENTS-CMU.md](https://github.com/AU-14/ColonialMarinesUniverse/blob/master/AGENTS-CMU.md#respect-cmu-boundaries) for the merge policy. Please add the missing tags, or let us know which hunks are purely mechanical and we'll take another look.

maertayn:
Saved reply, I noticed some parts were missing the tags, of CMU14 owned code outside of the zone, so an update wizden/rmc could lose it and silently revert it back.

PR #1986, opened 2026-09-21
Despoiler melee attack standardization and tailstab fix.

ILoveDespoiler:
## About the PR
Changed the damage of Despoilers melee attack to 18 Slash and 7 burn to 6 blunt, 6 slash, and 6 peirce, and 7 burn. as is standard with every other xenos melee attacks. also added 10 burn damage onto its tailstab, as previously it did 6.66... of each type, which means it did less damage than regular slash (20 compared to 25), which is lower than every other xeno, now it does 30, split between 3 brute types and Heat.

## Why / Balance
Normally Xeno claws do all three brute types, which I assume has something to do with armor resists. This standardizes despoilers brute damage types. As for the tailstab, every other xeno does more damage on tailstab compared to a slash, and now Despoiler does as well.

EDIT: after further testing, it appears that Despoiler is INTENDED to be doing 10Heat on tailstab, but is not. Therefore I am labeling that change as a BUGFIX and not a TWEAK

## Technical details
Changed the 18 Slash in despoilers melee to 6 of each brute type, as is standard. Added 10 Heat to the tailstab damage, which means it does 30 damage now, as is standard (see Drone, who does 25 brute on slash and 30 on tailstab)

## Test plan
I went into a localhost and used the tailstab to confirm it only does 20 damage. then I went back in after updating to confirm the extra Heat was being added properly. The claw thing was slightly harder to test, but the injury description thing seemed to indicate they were taking the blunt and peirce as well as the slash now.

## Requirements

**Changelog**
:cl:
- fix: Added 10 Heat damage to Despoilers tailstab, which was seemingly intended to apply when the target had no acid, but was non functional. Finishing Tail Stab now does 10 (Default) -> 20 -> 30 -> 40 Heat depending on lingering acid.
- fix: Fixed Despoilers slashes doing ONLY Slash compared to other xenos who do all 3 brute damage types.


ILoveDespoiler:
@maertayn I don't really agree with the balance label here... Unless Despoiler is *supposed* to be doing less damage than a Drone on tailstab and having its brute damage set up in a totally different way to every single other xeno

Honestly between this and what I've discovered about the lingering acid effects and caustic embrace, Despoiler might only be bad because large portions of its kit are completely nonfunctional

